PR #16, opened 2022-12-13
draft: Upload homework 2

oprisor-adrian:


course_2_homework/src/read_write_xml.cc
```diff
@@ -0,0 +1,52 @@
...
+	return file_content;
+}
+
+bool ReadWriteXML::set_file_path(std::string const& new_file_path){
+	if(exist(new_file_path)){
+		file_path = new_file_path;
+		return true;
+	}
+	throw std::invalid_argument(new_file_path + " invalide file path!" );
+	return false;
```
enachebogdan, line 36: you don't need to return, the function will exit after throwing anyway. nice that that you're using exceptions.

course_2_homework/src/read_write_xml.cc
```diff
@@ -0,0 +1,52 @@
...
+}
+
+bool ReadWriteXML::open_file(std::ifstream& input_file) const{
+	if(exist(file_path)){
+		input_file.close();
+		input_file.open(file_path, std::ifstream::in);
+		return true;
+	}
+	throw std::invalid_argument(file_path + " invalid file path!");
+	return false;
```
enachebogdan, line 51: same here, no need to return after throw.
since these functions throw errors, ideally you should call them in a try{} - catch() scope.

enachebogdan (changes requested):
Great job Adrian, code looks clean and well written and you implemented most of the required functionality. Most of my comments relate to minor issues, your homework looks very good overall.

course_2_homework/include/manage_store.h
```diff
@@ -0,0 +1,36 @@
+#ifndef MANAGE_STORE_INCLUDE_MANAGE_STORE_H_
```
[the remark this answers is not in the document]
enachebogdan: Cristi says Google guideline recommends #ifndef instead of #pragma, personal opinion is that you can use whichever but is good to know about both, so no need to change anything here

course_2_homework/include/product.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MANAGE_STORE_INCLUDE_PRODUCT_H_
+#define MANAGE_STORE_INCLUDE_PRODUCT_H_
+
+#include <iostream>
+#include <string>
+#include <tuple>
```
CristianIonutIonescu, line 6: you can get rid of this include

course_2_homework/include/product.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MANAGE_STORE_INCLUDE_PRODUCT_H_
+#define MANAGE_STORE_INCLUDE_PRODUCT_H_
+
+#include <iostream>
```
CristianIonutIonescu, line 4: you can get rid of this include here.
If you need it in source file, include it there
https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

course_2_homework/include/product.h
```diff
@@ -0,0 +1,38 @@
...
+  std::string GetName() const;
+  std::string GetCategory() const;
+  unsigned int GetQuantity() const;
+  float GetPrice() const;
+  void SetId(unsigned int const& id);
+  void SetName(std::string const& name);
+  void SetCategory(std::string const& category);
+  void SetQuantity(unsigned int const& quantity);
+  void SetPrice(float const& price);
+  void visit(void (*f)(unsigned int&,std::string&,std::string&,unsigned int&,float&));
```
CristianIonutIonescu, line 30: please check std::function<> object

course_2_homework/include/store_manager.h
```diff
@@ -0,0 +1,44 @@
+#ifndef STORE_INCLUDE_STORE_MANAGER_H_
+#define STORE_INCLUDE_STORE_MANAGER_H_
+
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "xml_reader_writer.h"
```
CristianIonutIonescu, line 8: same as above, if you don't use them in headers, don't include them

course_2_homework/include/xml_parser.h
```diff
@@ -0,0 +1,26 @@
+#ifndef STORE_MANAGER_INCLUDE_XML_PARSER_H_
+#define STORE_MANAGER_INCLUDE_XML_PARSER_H_
+
+#include <iostream>
```
CristianIonutIonescu, line 4: remove this

course_2_homework/src/xml_parser.cc
```diff
@@ -0,0 +1,85 @@
...
+}
+
+std::string XMLParser::GetInfo(std::string const& row){
+  unsigned int pos1 = row.find(">",0,1);
+  unsigned int pos2 = row.find("<",pos1,1);
+  return row.substr(pos1+1, pos2-pos1-1);
+}
+
+Product XMLParser::ConvertData(std::vector<std::string> const& unconverted_data){
+  unsigned int id = std::stoul(unconverted_data[0]);
```
CristianIonutIonescu, line 80: what if uncoverted_data size is less than 5?

course_2_homework/src/store_manager.cc
```diff
@@ -0,0 +1,188 @@
...
+  LoadData(path);
+}
+
+//copy constructor
+StoreManager::StoreManager(StoreManager const& other)
+    : no_of_products_(other.no_of_products_), data_(other.data_.begin(), other.data_.end()),
+      name_(other.name_){}
+
+//move constructor
+StoreManager::StoreManager(StoreManager&& other) noexcept
```
CristianIonutIonescu, line 23: re-read about move constructors, why do you initialise the members?

course_2_homework/src/store_manager.cc
```diff
@@ -0,0 +1,188 @@
...
+    : data_(std::vector<Product>()), no_of_products_(0), name_(std::string()){
+  no_of_products_ = other.no_of_products_;
+  name_ = other.name_;
+  data_.reserve(no_of_products_);
+
+  for(auto iterator : other.data_)
+    data_.push_back(iterator);
+
+  other.no_of_products_ = 0;
```
CristianIonutIonescu, line 32: we don't care about other data anymore, it should have been an empty shell

course_2_homework/src/store_manager.cc
```diff
@@ -0,0 +1,188 @@
...
+    data_.reserve(no_of_products_);
+
+    for(auto const& iterator : other.data_)
+	    data_.push_back(iterator);
+  }
+  return *this;
+}
+
+//move asignment constructor
+StoreManager& StoreManager::operator=(StoreManager&& other) noexcept{
```
CristianIonutIonescu, line 53: that's not how a move assigment operator should look

CristianIonutIonescu (changes requested):
Left some minor comments about include what you need and a question.
Most important remark:
Move constructors and assignment operators are not correct. Re-read about them and redo them

course_2_homework/src/xml_parser.cc
oprisor-adrian: I'm going to have an error because he's going to try to convert an element that doesn't exist. I can fix this issue by introducing a condition that checks the size of the vector.
`try{
   if(unconverted_data.size() == 5){
     // make conversion
   }
   throw std::invalid_argument("File contain incomplete product data set. Check it!");
}
catch(...){
  // do something
}`

course_2_homework/src/store_manager.cc
```diff
@@ -20,49 +24,39 @@ StoreManager::StoreManager(StoreManager const& other)
       name_(other.name_){}
 
 //move constructor
-StoreManager::StoreManager(StoreManager&& other) noexcept
-    : data_(std::vector<Product>()), no_of_products_(0), name_(std::string()){
+StoreManager::StoreManager(StoreManager&& other){ 
+  data_.clear();
```
CristianIonutIonescu, line 28: no need to clear by hand, C++ runtime will take care of it

course_2_homework/src/store_manager.cc
```diff
@@ -20,49 +24,39 @@ StoreManager::StoreManager(StoreManager const& other)
...
   no_of_products_ = other.no_of_products_;
   name_ = other.name_;
   data_.reserve(no_of_products_);
 
   for(auto iterator : other.data_)
```
CristianIonutIonescu, line 33: your container knows how to implement std::move, no need to move the objects ourselves (might confuse the compiler)

course_2_homework/src/store_manager.cc
```diff
@@ -20,49 +24,39 @@ StoreManager::StoreManager(StoreManager const& other)
...
+      data_.push_back(iterator);
   }
   return *this;
 }
 
 //move asignment constructor
-StoreManager& StoreManager::operator=(StoreManager&& other) noexcept{
+StoreManager& StoreManager::operator=(StoreManager&& other){
   if(this != &other){
     data_.clear();
```
CristianIonutIonescu, line 53: same comments as for move constructor

CristianIonutIonescu (commented):
Please fix latest comments and I'll approve afterwards

enachebogdan (approved):

PR #17, opened 2022-12-13
Lpitu tema2

Lys4nder:


product.cpp
```diff
@@ -0,0 +1,46 @@
+#include "store.hpp"
```
enachebogdan, line 1: did you mean to include product.hpp instead? don't think you need store class in here.

product.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+//Constructors
+Product::Product(string name_, double price_) : name(name_), price(price_) {}
+
+Product:://Copy constructor
+Product::Product(const Product& p) : name(p.name), price(p.price) 
+{ cout << "copy constructor called\n"; }
```
enachebogdan, line 8: even for one-liner functions, for readability is better to use {} on different lines and use indentation, like you used for the destructor for example. you can apply this for getters and setters as well.

product.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include <iostream>
+#include <string>
+using namespace std;
+
+class Product {
+    string name;
+    double price;
+    public:
```
enachebogdan, line 9: for better visibility, usually we indent public/private/protected differently from the members under them. You can either remove spacing for public: or add another tab to everything under the category.

product.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include <iostream>
+#include <string>
+using namespace std;
```
enachebogdan, line 4: usually we recommend against the using keyword for the entire namespace, instead write std::cout or whatever you need.

xmlClass.hpp
```diff
@@ -0,0 +1,44 @@
+#pragma once
+#include <iostream>
+#include <string>
+#include "rapidXML/rapidxml.hpp"
```
enachebogdan, line 4: good thing that you used an external parcer, it wouldn't be a problem if you implemented your own solution anyway.

store.cpp
```diff
@@ -0,0 +1,78 @@
...
+            {
+                product.setName("lapte");
+                product.setPrice(i+i/0.33);
+                products.push_back(product);
+            }
+    }
+
+    void Store::printProducts()
+    {
+        for (int i=0; i<products.size(); i++)
```
enachebogdan, line 73: try using range-based for loop or iterators to parce your vector

enachebogdan (changes requested):
Great job, code loos good, you implemented most of the required functionality. My comments are only for small issues, i'll have a second look once you've added the file parsing logic.

.gitignore
```diff
@@ -0,0 +1,4 @@
+/build
+/.vscode
+/rapidXML
```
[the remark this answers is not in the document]
Lys4nder: sure! i will make sure to remove the folder from .gitignore. Is is mandatory to modify the CMake file as well, as there are only hpp's, not cpp's in the library folder?

Lys4nder:
am dat push la versiunea completa

XMLObject.cpp
```diff
@@ -0,0 +1,28 @@
+#include "XMLObject.hpp"
+
+const std::vector<Product> XMLReader::read()
+{
+            std::ifstream file(fname);
+            std::string xmlString((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
+            
+            if (file)
+                std::cout << "file '" << fname << "' opened\n";
```
CristianIonutIonescu, line 9: if the file not found/cannot be opened why are we still trying to read from it?

CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.22)
+
+project(tema2)
+add_executable(tema2 main.cpp product.cpp store.cpp XMLObject.cpp)
+
+include_directories(/usr/include/rapidxml)
```
CristianIonutIonescu, line 6: might not be needed if
1. target_include_directories(point to correct interface)
2. you should include it after you checked for the presence of the library

CristianIonutIonescu (changes requested):
Mentioning here 2 things that I saw in multiple files:
1. [Order or declaring ](https://google.github.io/styleguide/cppguide.html#Declaration_Order)
2. ALL if/else/for/while statements must have the body clearly defined between { }even if it's one line:
```
if(..) {
} else {
}
for(...) {
}
//etc
```

Lys4nder:
done 😄

CristianIonutIonescu (approved):

enachebogdan (approved):

PR #19, opened 2022-12-14
Tema2

crissty21:


tema2/XmlParser.cpp
```diff
@@ -0,0 +1,45 @@
+#include "headers/XmlParser.h"
+#include <cstdlib>
+
+void XmlParser::readProductInfo(std::map<std::string, std::pair<std::string, std::string>> &productInfo)
+{
+    std::ifstream file(filePath);
+    std::string line;
+    if (!file.is_open())
+        std::cout << "file not opened";
```
enachebogdan, line 9: you could also return from function, otherwise it will try reading from a file that's not open

tema2/headers/XmlParser.h
```diff
@@ -0,0 +1,16 @@
+#include <iostream>
+#include <fstream>
```
enachebogdan, line 2: if you don't need a header in your .h files, it's good practice to include them in the .cpp file where you actually use it

enachebogdan (changes requested):
Good work Cristi, code looks clean and well written. I understand that this is just an initial draft, the homework requirement is to implement all "rule of five" functions, so I'm looking forward to see your update.

crissty21 (commented):
fixed everything

.vscode/settings.json
```diff
@@ -0,0 +1,52 @@
+{
```
CristianIonutIonescu, line 1: Add .vscode to ignore list

CristianIonutIonescu (approved):

crissty21:
this is the final form.
let me know if I need to modify something

CristianIonutIonescu (approved):

enachebogdan (approved):

PR #20, opened 2022-12-14
tema2DHoka added

radavid2017:


tema1DHoka/.gitignore
```diff
@@ -0,0 +1,2 @@
+.github
```
CristianIonutIonescu, line 1: You can have only one .gitignore in project root folder. This won't work

tema2DHoka/.gitignore
```diff
@@ -0,0 +1,2 @@
+build
```
CristianIonutIonescu, line 1: Same as above

tema2DHoka/file.cpp
```diff
@@ -0,0 +1,51 @@
...
+        while(getline(my_file, line)) {
+            vector<string> tokenized = tokenize(line, re);
+            string nume = tokenized[0];
+            int cant = stoi(tokenized[1]);
+            int pret = stoi(tokenized[2]);
+            store.addProduct(Product(nume, cant, pret));
+        }
+        cout << "S-a citit din fisier\n";
+    }
+    else cout << "Nu s-a deschis fisierul\n";
```
CristianIonutIonescu, line 30: Event though is correct add {} to else

tema2DHoka/file.cpp
```diff
@@ -0,0 +1,51 @@
...
+    my_file.close();
+}
+
+void StoreIO::write(Store& store) {
+    ofstream my_file(name);
+    if (my_file.is_open()) {
+        my_file.clear();
+        if (!store.isEmpty()) {
+        // for (auto itr = store.getData().begin(); itr < store.getData().end() ;itr++) {
+        //     my_file << itr->getNume() << " " << itr->getCantitate() << " " << itr->getPret() << endl;
```
CristianIonutIonescu, line 40: Is there a problem with iterators?

tema2DHoka/product.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+#include <iostream>
+using namespace std;
```
CristianIonutIonescu, line 3: No using namespace std

tema2DHoka/product.h
```diff
@@ -0,0 +1,42 @@
...
+
+class Product {
+    string nume;
```
CristianIonutIonescu, line 6: https://google.github.io/styleguide/cppguide.html#Declaration_Order
Check declaration order

tema2DHoka/product.h
```diff
@@ -0,0 +1,42 @@
...
+    int cant; // cantitate
+    int pret;
+    public:
+    Product(string& nume_produs, int& cantitate_produs, int& pret_produs) : nume(nume_produs), cant(cantitate_produs), pret(pret_produs) {
```
CristianIonutIonescu, line 10: No code written in header files.

Exception: inline functions, templates

tema2DHoka/store.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
+#include <iostream>
+#include <vector>
+#include "product.h"
+using namespace std;
```
CristianIonutIonescu, line 5: Same regarding std

CristianIonutIonescu (changes requested):
Thank you for the PRs. Glad to see that you used regex, STL algorithms and lambda functions.

I have left several comments that need to be addressed.
Also do NOT remove the codeowners files.

tema1DHoka/.gitignore
CristianIonutIonescu: Actually it works. Need to update my git knowledge

tema2DHoka/.gitignore
CristianIonutIonescu: Actually it works. Need to update my git knowledge

tema2DHoka/file.cpp
radavid2017: Yes, it was giving an error: terminate called after throwing an instance of '...' bad_alloc/length_error. But now, aparently it works fine. Maybe the problem was at something else where seems like I resolved it somehow.

tema2DHoka/product.h
```diff
@@ -0,0 +1,42 @@
...
+        // afisare in flux
+        os << "nume: " << product_out.getNume() << " | cant: " << product_out.getCantitate() << " | pret: " << product_out.getPret() << endl;
+        return os;
+    }
+    ~Product() {}
+    // getteri
+    string& getNume() {  return nume; }
+    int& getCantitate() {    return cant; }
+    int& getPret() { return pret; }
+    Product& getProduct() {  return *this; }
```
enachebogdan, line 36: What do you need this function for?

tema2DHoka/store.h
```diff
@@ -0,0 +1,61 @@
...
+
+    friend ostream& operator<<(ostream& os, Store& store_out) {
+        os << "\n-------- Store content --------\n";
+        for (auto itr = store_out.data.begin(); itr != store_out.data.end(); ++itr) {
+            os << *itr;
+        }
+        return os;
+    }
+    void display() {
+        for (auto itr = data.begin(); itr != data.end(); itr++) {
```
enachebogdan, line 56: In some places you use itr++, others ++itr, it's good to be consistent in general. ++itr is the better one (you can look up the differences, it's a good read)

tema2DHoka/product.h
radavid2017: Ah, actually I think it doesn't help me with anything..because the object itself is not private. I saw that method on stackoverflow.com and I thought it would help me to get the product through store. But it's useless actually.

tema2DHoka/store.h
```diff
@@ -0,0 +1,61 @@
...
+        return *this;
+    }
+    Store() {}
+    // MOVE CONSTRUCT
+    Store(Store&& other) noexcept {
+        cout << "Move constructor apelat\n";
+        data=other.getData();
+        // other.~Store();
+    }
+    Store* operator=(Store&& other) noexcept {
```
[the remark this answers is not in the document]
radavid2017: nagarrophoto
I saw this on course and I guessed that move construct is differentiated by that. That's the reason for doing it like this. Sorry if I missunderstood something here.

tema2DHoka/store.h
radavid2017: I changed it in some parts because I was debugging some problems on the iterations. Thanks for advice, I saw some informations about it and I think that if I am not using the initial value, it is more efficient to pre-increment.

CristianIonutIonescu (approved):

PR #21, opened 2022-12-14
First draft review

aronkis:


H2/include/Product.h
```diff
@@ -0,0 +1,23 @@
+#ifndef _PRODUCT_MANAGER_PRODUCT_H_
+#define _PRODUCT_MANAGER_PRODUCT_H_
+
+#include <string>
+
+class Product
```
CristianIonutIonescu, line 6: https://google.github.io/styleguide/cppguide.html#Declaration_Order

Declaration order:
public
protected
private

H2/.gitignore
```diff
@@ -0,0 +1,2 @@
+.vscode/
```
CristianIonutIonescu, line 1: Fairly sure that this doesn't work (unless github changes things since I last read about gitignore)

CristianIonutIonescu (commented):
Thank you for the PR.
Left some minor comments, nothing deal breaker.
Waiting to see the rest of the code

H2/.gitignore
CristianIonutIonescu: I stand corrected, you can have multiple gitignore files.

H2/include/ProductManager.h
```diff
@@ -0,0 +1,23 @@
...
+#define _PRODUCT_MANAGER_PRODUCT_MANAGER_H_
+
+#include "Product.h"
+#include <list>
+#include <iostream>
+
+class ProductManager
+{
+private:
+  std::list<ProductManager> productList;
```
enachebogdan, line 11: nothing wrong here, just wondering why you're using list instead of vector

H2/include/Product.h
```diff
@@ -0,0 +1,23 @@
...
+{
+private:
+  std::string name;
+  double price;
+public:
+  Product(std::string name, double price);
+  ~Product();
+  Product (const Product& other) = default;
+  Product (Product&& other) = default;
```
enachebogdan, line 15: having move/copy constructors as default is perfectly fine for your class, however for the homework purpose you should try implementing them.

H2/include/Product.h
```diff
@@ -0,0 +1,23 @@
...
+  Product& operator=(const Product& other);
+  Product& operator=(Product &&other);
```
enachebogdan, line 17: Since you're not declaring these assignment operators "= default", you'll have to implement them as well.

enachebogdan (changes requested):
Initial draft looks good so far, looking forward to seeing the rest of the implementation

H2/include/ProductManager.h
aronkis: because it is more efficient to remove elements from a list than from a vector

CristianIonutIonescu: > because it is more efficient to remove elements from a list than from a vector

[intrigued](https://tenor.com/en-GB/view/curiosity-curious-intrigued-interested-interesting-gif-14505707)

enachebogdan: sounds great, good call

H2/main.cc
```diff
@@ -0,0 +1,39 @@
...
+int main()
+{
+  XMLHandler Handler = XMLHandler("store.xml");
+  Handler.readFromFile();
+  std::vector<std::pair<std::string, double>> values = Handler.getValues();
+  //Handler.writeToFile();
+
+  for (auto it = values.begin(); it != values.end(); ++it)
+  {
+    std::cout << (*it).first << " " << (*it).second << std::endl;
```
CristianIonutIonescu: why not using it->first and it->second ?

H2/main.cc
```diff
@@ -0,0 +1,39 @@
...
+  Product monitor = Product("Monitor", 330.23);
+
+
+  values.push_back(std::make_pair(mouse.getName(), mouse.getPrice()));
+  values.push_back(std::make_pair(keyboard.getName(), keyboard.getPrice()));
+  values.push_back(std::make_pair(monitor.getName(), monitor.getPrice()));
+
+  for (auto it = values.begin(); it != values.end(); ++it)
+  {
+    std::cout << (*it).first << " " << (*it).second << std::endl;
```
CristianIonutIonescu: same as above.
Bonus read about const interators and when they are used.

H2/include/ProductManager.h
CristianIonutIonescu: following up on this:
shouldn't be a` list<Product>` instead of `list<ProductManager>` ?

Also, to not add another comment, access order again

H2/main.cc
```diff
@@ -0,0 +1,39 @@
...
+#include <vector>
+#include <utility>
+#include <iostream>
+
+
+int main()
+{
+  XMLHandler Handler = XMLHandler("store.xml");
+  Handler.readFromFile();
+  std::vector<std::pair<std::string, double>> values = Handler.getValues();
```
CristianIonutIonescu: what operation is performed under the hood?
Copy or move? And why?

H2/main.cc
```diff
@@ -0,0 +1,39 @@
...
+  {
+    std::cout << (*it).first << " " << (*it).second << std::endl;
+  }
+  
+  Product mouse = Product("Mouse", 150.0);
+  Product keyboard = Product("Keyboard", 175.76);
+  Product monitor = Product("Monitor", 330.23);
+
+
+  values.push_back(std::make_pair(mouse.getName(), mouse.getPrice()));
```
CristianIonutIonescu: code works ok as it.
Questions:
1. What operator/constructor is called under the hood? Copy or more?
2. Read about `emplace_back` method and how it would have been used here?

H2/.gitignore
```diff
@@ -0,0 +1,6 @@
...
+build/
+CMakeFiles/
```
CristianIonutIonescu, line 3: this tells me you have built your application wrong :D

H2/src/Product.cc
```diff
@@ -0,0 +1,55 @@
+#include "Product.h"
+#include <iostream>
+#include <utility>
+
+Product::Product(std::string name, double price)
```
CristianIonutIonescu, line 5: read about Constructor initializer lists.
Same applies to other constructors

H2/src/XMLHandler.cc
```diff
@@ -0,0 +1,103 @@
...
+  path_to_file_ = std::move(other.path_to_file_);
+  file_content_ = std::move(other.file_content_);
+  return *this;
+}
+
+std::vector<std::pair<std::string, double>> XMLHandler::readFromFile()
+{
+  std::cout << "XMLHandler::readFromFile called." << std::endl;
+  std::ifstream xmlFile;
+  std::string line;
```
CristianIonutIonescu, line 68: why declaring them (line,values, tempPair) here when you might not even use them?
Restrict the lifetime of your objects to the minimum.
Also why do you need values and tempPair objects? can you store them directly in file_content?

H2/src/XMLHandler.cc
```diff
@@ -0,0 +1,103 @@
...
+  {
+    std::cout << "Cannot open the file." << std::endl;
+  }
+}
+
+void XMLHandler::writeToFile()
+{
+  std::cout << "XMLHandler::writeToFile called." << std::endl;
+  std::ofstream xmlFile;
+  xmlFile.open(path_to_file_, std::ofstream::out | std::ofstream::trunc);
```
CristianIonutIonescu, line 103: what if xmlFile.open fails?

H2/include/XMLHandler.h
```diff
@@ -0,0 +1,30 @@
...
+#include <utility>
+#include <list>
+
+class XMLHandler
+{
+public:
+  XMLHandler(std::string file);
+  XMLHandler(std::string file, std::vector<std::pair<std::string, double>> file_content);
+  XMLHandler(std::string file, std::list<Product> file_content);
+  ~XMLHandler() {}
```
CristianIonutIonescu, line 16: = default

CristianIonutIonescu (commented):
Most of the comments are minor and don't influence the outcome of homework.
Just extra stuff

H2/main.cc
aronkis: great thing I asked about them and forgot to use them :)

H2/main.cc
aronkis: The move operation is performed, because we do not use the Handler anymore so the compiler might discard it after the move.

CristianIonutIonescu (approved):

H2/main.cc
aronkis: 1. The move operator, because after we created the copy the compiler frees up the memory immediately.
2. The `emplace_back` method does not create a temporary vector, like `push_back`, instead it appends the values directly into the vector, which means its faster and memory efficient.

enachebogdan (approved):

PR #27, opened 2023-02-26
Upload homework 3

oprisor-adrian:


course_3_homework/src/file_wrapper.cc
```diff
@@ -0,0 +1,57 @@
+#include "../include/file_wrapper.h"
+
+#include <iostream>
+
+FileWrapper::FileWrapper(): stream_(nullptr){}
```
enachebogdan, line 5: Your FileWrapper class doesn't quite respect the RAII principle.
You would need to aquire the FILE* resource in the constructor, you could also send a parametr for opening the file for read/write. You keep the FILE* resource for the entire lifetime of the FileWrapper, no need to open/close it again. You close the stream on the destructor.

This is all that RAII does actually, you create a FileWrapper object on stack, so when it goes out of scope the destructor will automatically call the fclose. This protects you from the  multiple possible scenarios (when handling the FILE* directly) where the code doesn't actually call fclose and you have dangling file descriptor

course_3_homework/src/file_wrapper.cc
```diff
@@ -0,0 +1,57 @@
...
+    stream_ = nullptr;
+  }
+  return stream_;
+}
+
+std::FILE* FileWrapper::Get() const {
+  return stream_;
+}
+
+void FileWrapper::FileOpener(const char* path) {
```
enachebogdan, line 50: Just a suggestion on naming functions, they should have action names, not objects. For example, in your case it should be FileWrapper::FileOpen(...)

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+    UniquePointer(UniquePointer&& other) noexcept;
+    UniquePointer& operator=(const UniquePointer& other) = delete;
+    UniquePointer& operator=(UniquePointer&& other) noexcept;
+    ~UniquePointer();
+    explicit operator bool() const;
+    data_type* operator->() const;
+    data_type& operator*() const;
+    data_type& operator[](std::size_t index) const;
+    data_type* Release();
+    void Reset(data_type* pointer = 0);
```
enachebogdan, line 29: use nullptr instead of 0.

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+    : data_(std::move(data)), deleter_(std::move(deleter)){}	
+
+template<class data_type, class deleter_type>
+UniquePointer<data_type, deleter_type>::UniquePointer(data_type* data[])
+    : data_(data){}
+
+template<class data_type, class deleter_type>
+UniquePointer<data_type, deleter_type>::UniquePointer(
+    UniquePointer&& other) noexcept{
+  this->Reset(other.Release());  // replace `*this` managed object and release `other'
```
enachebogdan, line 58: nice use of the functions here

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+  if(Get() != nullptr){
+    deleter_(data_);
+  }
+}
+
+template<class data_type, class deleter_type>
+UniquePointer<data_type, deleter_type>::operator bool() const {
+  if(Get() != nullptr)
+    return true;
+  return false;
```
enachebogdan, line 86: you could just use:
return data_ != nullptr, you get rid of a function call + an if statement

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+   this->Reset(other.Release());
+   this->deleter_ = std::move(other.deleter_);
+   // this->Swap(other);  // transfer ownership from 'other' to '*this'
+  }
+  return *this;
+}
+
+template<class data_type, class deleter_type>
+UniquePointer<data_type, deleter_type>::~UniquePointer(){
+  if(Get() != nullptr){
```
enachebogdan, line 77: should compare the data_ directly instead of using Get(), you get rid of an unnecessary function call.

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+template<class data_type, class deleter_type>
+UniquePointer<data_type, deleter_type>::operator bool() const {
+  if(Get() != nullptr)
+    return true;
+  return false;
+}
+
+template<class data_type, class deleter_type>
+data_type* UniquePointer<data_type, deleter_type>::operator->() const {
+  return Get();
```
enachebogdan, line 91: You can see already I'm not a big fan of calling Get() when you don't have to, I'd get rid of it all internal functions 😄

course_3_homework/include/unique_ptr.h
```diff
@@ -0,0 +1,162 @@
...
+  data_ = pointer;
+  if(old_pointer){
+    GetDeleter()(old_pointer);
+  }
+}
+
+template<class data_type, class deleter_type>
+void UniquePointer<data_type, deleter_type>::Swap(UniquePointer& other) {
+  // swap data_
+  data_type* aux_data = std::move(other.data_);
```
enachebogdan, line 124: Did you try using std::swap?

course_3_homework/include/shared_ptr.h
```diff
@@ -0,0 +1,164 @@
...
+      delete data_;
+      delete counter_;
+    }
+  }
+}
+
+template<class data_type>
+void SharedPointer<data_type>::Reset(){
+  if(Get() != nullptr){
+    SharedPointer().Swap(*this);
```
enachebogdan, line 86: I'd say it would be clear to just reset the values instead of creating a new empty shared_ptr and swap with *this

course_3_homework/include/shared_ptr.h
```diff
@@ -0,0 +1,164 @@
...
+    (*counter_)--;
+  }
+}
+
+template<class data_type>
+template<class shared_type>
+void SharedPointer<data_type>::Reset(shared_type* data){
+  if(Get() != nullptr){
```
enachebogdan, line 94: Same comment with Get() as for unique_ptr functions

course_3_homework/include/shared_ptr.h
```diff
@@ -0,0 +1,164 @@
...
+    SharedPointer(data).Swap(*this);
+  }
+}
+
+template<class data_type>
+void SharedPointer<data_type>::Swap(SharedPointer& other){
+  // move data_
+  data_type* aux_data = std::move(data_);
```
enachebogdan, line 102: similar to unique, did you try std::swap?

enachebogdan (commented):
PR looks good, some changes here and there but overall code is fine. I'll have another look after your next commits

enachebogdan (approved):

PR #28, opened 2023-02-27
DHoka tema 3

radavid2017:


tema3DHoka/shared.h
```diff
@@ -0,0 +1,165 @@
...
+
+#define _SHARED_PTR_
+
+#include <iostream>
+
+template <class T>
+class shared_ptr {
+    public:
+        shared_ptr(); // constructor default
+        shared_ptr(T pointer_object); // constructor
```
enachebogdan, line 11: Would be better to create from a T* instead of T.

tema3DHoka/shared.h
```diff
@@ -0,0 +1,165 @@
...
+        int* count = nullptr;
+};
+
+template <class T>
+shared_ptr<T>::shared_ptr() {
+    data = nullptr;
+}
+
+template <class T>
+shared_ptr<T>::shared_ptr(T pointer_object) {
```
enachebogdan, line 41: Use member initializer list aftet you change to T* type

tema3DHoka/shared.h
```diff
@@ -0,0 +1,165 @@
...
+
+template <class T>
+T* shared_ptr<T>::operator->() const {
+    return data;
+}
+
+template <class T>
+void shared_ptr<T>::clean_up() {
+    if (count != nullptr) {
+        (*count)--;
```
enachebogdan, line 79: Are there any cases where *count can be 0 here?

clean_up() is not a function exposed in shared_pointer interface in std::, so not sure why you need it.

Another issue, this is a public funtion so somebody can call clean_up on a shared ptr, it removes data and count. You do not set count to nullptr here, you just delete it, so it will generate undefined behaviour if  call the function again(directly or from some other function).

tema3DHoka/shared.h
```diff
@@ -0,0 +1,165 @@
...
+    return *this;
+}
+
+template <class T>
+const shared_ptr<T>& shared_ptr<T>::get() const {
+    return data;
+}
+
+template <class T>
+void shared_ptr<T>::set(shared_ptr& other) {
```
enachebogdan, line 126: set is also not part of the shared_ptr from std::, so you don't really need it. It's more or less similar to assignment operatror.

tema3DHoka/shared.h
```diff
@@ -0,0 +1,165 @@
...
+}
+
+template <class T>
+bool shared_ptr<T>::unique() const noexcept {
+    return (*count) == 1;
+}
+
+template <class T>
+shared_ptr<T>::operator bool() const noexcept {
+    return data == nullptr;
```
enachebogdan, line 162: operator should check if data is not null

tema3DHoka/unique.h
```diff
@@ -0,0 +1,125 @@
...
+        T* release(); // eliberare pointer
+        void reset(T* new_pointer); // overwrite cu noul pointer
+        void swap(unic_ptr& other); // interschimbare cu other
+        
+    private:
+        T* pointer = nullptr;
+};
+
+template <class T>
+unic_ptr<T>::unic_ptr() {
```
enachebogdan, line 40: You should use member initializer list for all your constructors when possible.

tema3DHoka/unique.h
```diff
@@ -0,0 +1,125 @@
...
+            os << _unic_ptr_out.get() << std::endl;
+            return os;
+        }
+        unic_ptr(); // constructor default
+        unic_ptr(T* pointer_object); // constructor
+        unic_ptr(const unic_ptr& other) = delete; // copy construct dezactivat
+        unic_ptr(unic_ptr&& right); // move construct
+        template <class U>
+        unic_ptr(unic_ptr<U>&& right); // move construct v2
+        unic_ptr& operator=(unic_ptr& _unic_ptr); // move assignment
```
enachebogdan, line 22: for move assignment you should have (unic_ptr&& _unic_ptr) param, this is just another form of copy assignment.

tema3DHoka/unique.h
```diff
@@ -0,0 +1,125 @@
...
+        T& operator*();
+        T* operator->();
+        ~unic_ptr(); // destructor
+        /* Observers */
+        const unic_ptr& get() const; // getter
+        operator bool(); // verifica pointer == nullptr
+        void set(unic_ptr& _unic_ptr_nou); // setter
```
enachebogdan, line 29: not needed, same case as for shared_ptr

tema3DHoka/unique.h
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <class T>
+unic_ptr<T>::~unic_ptr() {
+    this->release();
+    std::cout<<"unic_ptr destructor apelat\n";
+}
+
+template <class T>
+unic_ptr<T>& unic_ptr<T>::operator=(unic_ptr& _unic_ptr) {
```
enachebogdan, line 70: should use a better naming for the parameter, _unic_ptr is confusing here, maybe "other" or "right" as you used before.

enachebogdan (commented):
Good work David, with some changes here and there we are good to go.

tema3DHoka/filewrapper.cpp
```diff
@@ -0,0 +1,40 @@
+#include "filewrapper.h"
+
+FileWrapper::FileWrapper(std::string const& name) {
```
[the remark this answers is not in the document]
radavid2017: I understand. The class acomplish succesfully the RAII principle now. Thank you for the given support!

tema3DHoka/shared.h
radavid2017: I modified it. Now it's T* implemented as the parameter of the constructor.

tema3DHoka/shared.h
radavid2017: I deleted the clean_up method, realizing it is unneeded.

tema3DHoka/shared.h
radavid2017: Yes, I was mistaken there. Now I changed the operator to "!=".

tema3DHoka/unique.h
radavid2017: I modified the method, and in the main.cpp I wrote the move assignment with the library called move() which is found in the std library.

tema3DHoka/unique.h
radavid2017: Yes, I deleted it here also.

tema3DHoka/unique.h
radavid2017: In final version, I chose to rename the parameter with "right".

tema3DHoka/shared.h
```diff
@@ -0,0 +1,138 @@
...
+
+template <class T>
+T* shared_ptr<T>::operator->() const {
+    return data;
+}
+
+template <class T>
+shared_ptr<T>::~shared_ptr() {
+    this->data = nullptr;
+    this->count = nullptr;
```
enachebogdan, line 79: Now that you don't have the cleanup function anymore, you still need to decrease the count, then check if count has reached to 0 in order to delete "data" and "count" pointers. Same thing with the reset function

tema3DHoka/shared.h
```diff
@@ -0,0 +1,138 @@
...
+        (*count)++;
+    }
+    std::cout<<"shared_ptr copy assignment apelat\n";
+    return *this;
+}
+
+template <class T>
+shared_ptr<T>& shared_ptr<T>::operator=(shared_ptr&& other) {
+    data = other.data;
+    count = other.count;
```
enachebogdan, line 97: easier way to do this, use std::move for both data/count

tema3DHoka/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+int main() {
+
+    std::cout<<"Folosire unic pointers\n";
+    unic_ptr<Rectangle> up1(new Rectangle(10, 5));
+    std::cout<<"unicPointer1->area(): " << up1->area() << std::endl;
+    unic_ptr<Rectangle> up2;
+    up2 = std::move(up1); // move assignment apelat
+    std::cout<<"unicPointer2->area(): " << up2->area() << std::endl;
+    // testare bool operator up1
+    if (up1) std::cout<<"up1 INCA MAI RETINE ADRESA\n";
```
enachebogdan, line 29: even for small if/else instructions, try to use {}, it's good to be consistent and it improves readability.

enachebogdan (commented):
Good job fixing the issues David, I've added few more minor comments, some things that I've missed the first time.

tema3DHoka/shared.h
radavid2017: I think I understood better this time how the destructor and reset method work. If there are any issues, please tell me.

tema3DHoka/shared.h
```diff
@@ -0,0 +1,138 @@
...
+
+template <class T>
+shared_ptr<T>::~shared_ptr() {
+    this->data = nullptr;
+    this->count = nullptr;
+}
+
+template <class T>
+shared_ptr<T>& shared_ptr<T>::operator=(const shared_ptr& other) {
+    data = other.data;
```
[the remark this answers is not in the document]
radavid2017: I put that into the code, and also added the "&& this-> data != nullptr" when I assign to a null pointer, avoiding the segmentation fault error.

tema3DHoka/shared.h
radavid2017: Thanks for the advice, I followed it up.

enachebogdan (approved):

PR #29, opened 2023-02-28
lpitu_tema3

Lys4nder:


main.cpp
```diff
@@ -0,0 +1,13 @@
+#include "unique.hpp"
+#include "shared.hpp"
+#include <string>
+
+int main()
+{
+    uniquePtr<int> uptr(new int(16), [](int* p){});
```
enachebogdan, line 7: the lambda that you're passing as the "deleter" parameter does nothing, so any time you call the deleter it will not actually call the delete on your pointer. If you don't pass anything, the default parameter should take care of things.

enachebogdan (commented):
Looking very good so far, only minor comments from my end. I'll have another look on the next commits.

main.cpp
```diff
@@ -2,12 +2,14 @@
...
 #include "files.hpp"
 #include <string>
+#include <memory>
 
 int main()
 {
     uniquePtr<int> uptr(new int(16), [](int* p){});
     std::cout << *uptr << std::endl;
 
+
```
enachebogdan, line 12: Your latest commit doesn't change anything in the lambda function, it just adds a new empty line. Maybe it was some problem with your commit? If you need help with the fix or if my comments are not very clear feel free to ping me.

enachebogdan (approved):

enachebogdan (approved):

PR #30, opened 2023-02-28
hw3 draft

914-Mereu-Bianca:


enachebogdan (commented):
Code looks good, clean, well written. I look forward to seeing the complete assignment.

pointers/shared_ptr.h
```diff
@@ -0,0 +1,145 @@
...
+SharedPointer<T>::SharedPointer(const SharedPointer &other) noexcept : pointer(std::move(other.pointer)), 
+counter(std::move(other.counter)) { (*counter)++; }
+
+template <class T>
+SharedPointer<T> &SharedPointer<T>::operator=(const SharedPointer &other) noexcept {
+    if(this != &other) {    
+        pointer = other.pointer;
+        counter = other.counter;
+        (*counter)++;
+    }
```
enachebogdan, line 65: If we are no longer pointing to that old data, shouldn't we decrease the count before assigning it the new one?
Also, if the decreased count is now 0, we should release the memory (similar to what you did for destructor)

pointers/shared_ptr.h
```diff
@@ -0,0 +1,145 @@
...
+
+
+template <class T>
+SharedPointer<T>::SharedPointer() noexcept : pointer(nullptr), counter(new size_t(1)) {}
+
+template <class T>
+SharedPointer<T>::SharedPointer(nullptr_t) noexcept : pointer(nullptr), counter(new size_t(1)) {}
+
+template <class T>
+SharedPointer<T>::SharedPointer(T *new_pointer) noexcept : pointer(std::move(new_pointer)),
```
enachebogdan, line 52: Any particular reason for using std::move here?

pointers/shared_ptr.h
```diff
@@ -0,0 +1,145 @@
...
+
+template <class T>
+SharedPointer<T>::SharedPointer(nullptr_t) noexcept : pointer(nullptr), counter(new size_t(1)) {}
+
+template <class T>
+SharedPointer<T>::SharedPointer(T *new_pointer) noexcept : pointer(std::move(new_pointer)), 
+counter(new size_t(1)) {}
+
+template <class T>
+SharedPointer<T>::SharedPointer(const SharedPointer &other) noexcept : pointer(std::move(other.pointer)),
```
enachebogdan, line 56: Same question regarding the use of  std::move

filewrapper/filewrapper.h
```diff
@@ -0,0 +1,31 @@
...
+    public:
+    FileWrapper(const char* _file_name, const char* _mode);
+    ~FileWrapper();
+
+    private:
+    FILE *file;
+    std::string mode;
+};
+
+FileWrapper::FileWrapper(const char* _file_name, const char* _mode) {
```
enachebogdan, line 17: Normally I would complain that you should implement functions in a .cpp file, but for such small class it's fine.

main.cpp
```diff
@@ -1,11 +1,21 @@
 #include <iostream>
 #include "pointers/unique_ptr.h"
+#include "pointers/shared_ptr.h"
 
 int main(){
     std::string _file = "/home/bianca/2022/file.txt";
```
enachebogdan, line 6: I assume you wanted to create a FileWrapper here but you forgot 😄 ?

enachebogdan (commented):
Nice work, RAII looks good, smart pointers as well, I just added  few minor comments.

pointers/shared_ptr.h
914-Mereu-Bianca: Yes, I should've checked and edited the old data too, I'll add the changes

pointers/shared_ptr.h
914-Mereu-Bianca: Well I think that after the call of the function std::move we will lose anything from 'new_pointer', but it will remain in 'pointer'. I shouldn't modify the vaiable 'new_pointer'?

pointers/shared_ptr.h
914-Mereu-Bianca: I understand now, here I didn't have to use this function because it is a copy constructor, I'll edit it!

main.cpp
914-Mereu-Bianca: Yes, but I will finish the FileWrapper class first, then I will continue here!:)

pointers/shared_ptr.h
enachebogdan: First of all, new_pointer(without the *) is just an address to a memory zone, you just want to copy that address into your pointer, you don't need to move the data from one memory zone to another.

Secondly, std::move will just try to static_cast new_pointer to rvalue type, after that the appropriate "=" operator is called. Since the type of pointer/new_pointer is raw pointer, what it ends up doing is a simple pointer=new_pointer(will not do anything to new_pointer). If you would want to move the actual data, you would have to call std::move(*new_pointer).

pointers/shared_ptr.h
enachebogdan: Here you have a copy contructor, so you would not want to alter the "other" anyway, you are even sending it as a const.

enachebogdan (approved):

PR #32, opened 2023-03-01
Tema 3

andreea-bumbac:


homework3/include/shared_ptr.h
```diff
@@ -0,0 +1,98 @@
...
+    // getter numar de referinte
+    int getCount() const;
+
+    // reseteaza pointer-ul
+    void reset(T* pointer_);
+
+private:
+    T* pointer;
+    int* count;
+};
```
enachebogdan, line 30: You are missing some of the required functions for this assignment(swap, get, bool operator, etc..). Check the link that Cristi sent to see the full list.

homework3/include/shared_ptr.h
```diff
@@ -0,0 +1,98 @@
...
+
+// getter pentru numarul de referinte
+template<class T>
+inline int SharedPointer<T>::getCount() const {
+    return *count;
+};
+
+template<class T>
+inline void SharedPointer<T>::reset(T* pointer_) {
+    delete pointer;
```
enachebogdan, line 72: You should first check if pointer != pointer_... you wouldn't want to do a delete if that's the case.
Your instance of shared pointer might not be the only one holding the pointer data(and count). So you should't delete the pointer/count unless you are the last shared pointer that holds the data.

enachebogdan: I just noticed, you did things right for the = operator, this should be similar.

enachebogdan (commented):
Good PR overall, code looks nice and clean. Looking forward to seeing the rest of the implementation.

PR #33, opened 2023-03-01
Akis main v2

aronkis:


H3/Pointers/include/SharedPointer.h
```diff
@@ -0,0 +1,163 @@
...
+    dummy_pointer_ = other.dummy_pointer_;
+    count_ = other.count_;
+    other.dummy_pointer_ = nullptr;
+    other.count_ = new size_t(1);
+  }
+
+  // Modifiers
+  void reset() noexcept
+  {
+    SharedPointer().swap(*this);
```
enachebogdan, line 75: I would rather set up the values manually than doing copy-and-swap to a new object(just for readability).

H3/Pointers/include/SharedPointer.h
```diff
@@ -0,0 +1,163 @@
...
+
+  void reset(T* pointer) noexcept
+  {
+    SharedPointer(pointer).swap(*this);
+  }
+
+  void swap(SharedPointer& other) noexcept
+  {
+
+    T* dummy_temp_ = std::move(dummy_pointer_);
```
enachebogdan, line 86: did you try using std::swap?

H3/Pointers/include/SharedPointer.h
```diff
@@ -0,0 +1,163 @@
...
+    return *this->use_count == 1;
+  }
+
+  explicit operator bool() const noexcept
+  {
+    return get() != nullptr;
+  }
+
+  // ???
+  bool owner_before(const SharedPointer<T>& other) const noexcept
```
enachebogdan, line 138: no need to worry about this, I'd say we can skipp it.

H3/Pointers/include/UniquePointer.h
```diff
@@ -0,0 +1,156 @@
...
+  UniquePointer& operator=(UniquePointer<U,E>&& other)
+  {
+    dummy_pointer_ = other.release();
+    deleter_ = std::forward<Deleter>(other.get_deleter());
+  }
+
+  // Modifiers
+  void reset(T* pointer) noexcept
+  {
+    deleter_(dummy_pointer_);
```
enachebogdan, line 74: You should check if dummy_pointer == pointer, otherwise deleting it will cause all sort of problems.

enachebogdan (commented):
PR looks very good, on point with the RAII on file handling, smart pointers look good. Nice job!

enachebogdan (approved):

PR #34, opened 2023-03-06
General Store first draft.

aronkis:


H4/include/Client/Client.h
```diff
@@ -0,0 +1,32 @@
...
+{
+public:
+  Client() = default;
+  explicit Client(int age, std::string name): age_(age), name_(name) {}
+  virtual ~Client() = default;
+  Client(const Client& other);
+  Client(Client&& other);
+  Client& operator=(const Client& other);
+  Client& operator=(Client&& other);
+  bool operator==(const Client& other);
```
marisvali, line 20: Don't implement these unless needed.
https://en.cppreference.com/w/cpp/language/rule_of_three
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero
We are in the 'rule of zero' case.

This might be confusing as we asked you to implement these in the previous two homeworks. Homework 2 was a practice run specifically for implementing constructors/assignment/destructors. Homework 3 dealt with RAII and classes which are owners of raw pointers. So for these two assignments you had to implement these operations. From now on, implement only what is necessary.

Quote from the first link: "Classes that have custom destructors, copy/move constructors or copy/move assignment operators should deal exclusively with ownership (which follows from the Single Responsibility Principle). Other classes should not have custom destructors, copy/move constructors or copy/move assignment operators."

This applies to all classes in this assignment, follow the rule of zero if you can.

H4/include/Client/ClientManager.h
```diff
@@ -0,0 +1,25 @@
...
+  ClientManager(ClientManager&& other);
+  ClientManager& operator=(const ClientManager& other);
+  ClientManager& operator=(ClientManager&& other);
+  void removeClient(Client client);
+  void addClient(Client client);
+  std::list<Client> getClients();
+
+protected:
+  std::list<Client> clients_;
+};
```
marisvali, line 22: The client manager doesn't add any value. The class just acts as a list and could be replaced with a typedef:

typedef std::list<Client> ClientManager;

The class holds Client objects instead of pointers, so polymorphism will never be used. When you call addClient(Client client) and you send a PremiumClient object, you will get object slicing (https://en.wikipedia.org/wiki/Object_slicing). If you decide to keep a manager after you refactor the project and you want to keep a list of pointers, use a list of smart pointers if the manager owns the objects.

H4/src/Client/Client.cc
```diff
@@ -0,0 +1,77 @@
...
+
+double Client::calculateCheckOut()
+{
+  std::list<Product*>::iterator it;
+  double total = 0.0;
+  
+  for (it = cart_.begin(); it != cart_.end(); it++)
+  {
+    total += (*it)->getPrice();
+  }
```
marisvali, line 51: There's nothing wrong here. I just wanted to point out you can use the ranged-based for loop. I'm not sure if you know about it. I love it, you might like it too:

for (auto product: cart_)
{
    total += product->getPrice();
}

https://en.cppreference.com/w/cpp/language/range-for

marisvali (commented):
The program implements the following story: "There are some products. A client can be made to take one of a product, one by one. The total cost of the products for each client can be computed. The products can be added to a client manager then displayed as a list. The clients can also be added to a manager then displayed as a list."

The homework asked for something else. Here is an example of how the story should read: "There is a store. The store has an inventory (some products) and a balance (some money). Clients can get items from the store. When they do, the store earns money and the store's inventory is updated."

There is a hierarchy of products, but the hierarchy isn't used at all in main. The project doesn't show polymorphism or SOLID in action.

"Extend(or implement from scratch) a general store"
"increasing income"
"Multiple types of products" -> they should be used in main, currently they are implemented and never used. 'Product water("Water", 1, 10)' is used when the class 'NonAlcoholic' is available.
"customer enters shop, ask for a product"

My suggestion: rethink all the classes and connections between them. I'll post more details on Slack this evening.

H4/src/printer.cc
```diff
@@ -0,0 +1,21 @@
+#include "../include/printer.h"
+
+void Printer::print(Client& client) const
```
CristianIonutIonescu, line 3: [optional]mark them as const

H4/include/Product/Drink/alcoholic.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "beverage.h"
+
+class Alcoholic : public Beverage
+{
+public:
+  Alcoholic(std::string name, double price) 
+    : Beverage(name, price) {}
+private:
+  bool restricted_ = true;
```
CristianIonutIonescu, line 12: same as discount, maybe add it in beverage?

H4/include/Client/premium_client.h
```diff
@@ -0,0 +1,17 @@
...
+
+class PremiumClient : public Client
+{
+public:
+  PremiumClient(std::string name, int age, double balance)
+    : Client(name, age, balance) {}
+  bool AddToCart(Product* product, Store* store);
+  double GetTotal();
+private:
+  double discount_ = 0.25;
```
CristianIonutIonescu, line 14: why don't you have this as protected in base class, equal with 0 and in your derived class set to whatever value you want? this way you wouldn't need to override the GetTotal

H4/include/Client/premium_client.h
```diff
@@ -0,0 +1,17 @@
...
+#define _GENERAL_STORE_PREMIUM_CLIENT_H_
+
+#include "client.h"
+
+class PremiumClient : public Client
+{
+public:
+  PremiumClient(std::string name, int age, double balance)
+    : Client(name, age, balance) {}
+  bool AddToCart(Product* product, Store* store);
```
CristianIonutIonescu, line 11: wherever you use pointers, take a step back and think about the ownership (if it has ownership or just "visiting" and how many owners).

H4/include/Client/client.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _GENERAL_STORE_CLIENT_H_
+#define _GENERAL_STORE_CLIENT_H_
+
+#include <string>
+#include <list>
+#include <utility>
+#include "../include/Product/product.h"
```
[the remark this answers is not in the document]
aronkis: I tried but it didn't compile afterwards.

CristianIonutIonescu (approved):
Although there are many things that can be addressed, I consider that the majority of the requirements were met on a level enough for me to consider the homework done.

H5/include/Client/client.h
```diff
@@ -0,0 +1,36 @@
...
+
+using Cart = std::list<std::pair<std::shared_ptr<Product>, int>>;
+
+class Client
+{
+public:
+  Client(std::string name, int age, double balance, double discount = 0);
+  virtual ~Client() = default;
+  inline double GetBalance() const { return balance_; }
+  inline Cart GetCart() const {return cart_; }
```
CristianIonutIonescu, line 21: Return a const reference to avoid unnecessary copies.

H5/include/Client/client.h
```diff
@@ -0,0 +1,36 @@
...
+  inline std::string GetName() const { return name_; }
+  Cart::iterator FindInCart(std::shared_ptr<Product> product);
+  double GetTotal();
+  void Pay(double total);
+  virtual void AddToCart(std::shared_ptr<Product> product, std::shared_ptr<Store> store);
```
CristianIonutIonescu, line 26: Do you add a new owner to the store? If not you can pass just a raw pointer to visit it

H5/include/printer.h
```diff
@@ -0,0 +1,27 @@
+#ifndef _GENERAL_STORE_PRINTER_H_
+#define _GENERAL_STORE_PRINTER_H_
+
+#include "../include/Client/client.h"
+#include "../include/Product/product.h"
```
CristianIonutIonescu, line 5: You don't use the product here. You can forward declare the client and store as well

H5/include/printer.h
```diff
@@ -0,0 +1,27 @@
...
+#include "../include/store.h"
+
+class ClientPrinter
+{
+public:
+  virtual void print(const std::shared_ptr<Client> client) const = 0;
```
CristianIonutIonescu, line 11: Printer won't own the client, pass raw pointer. Same for store

H5/include/store.h
```diff
@@ -0,0 +1,31 @@
...
+{
+public:
+  Store(std::string name, double balance);
+  Inventory::iterator FindInStore(std::shared_ptr<Product> product);
+  void AddToStore(std::shared_ptr<Product> product, int quantity = 1);
+  void RemoveFromStore(std::shared_ptr<Product> product);
+  void PrintInventory();
+  void CheckOut(std::shared_ptr<Client> client);
+  inline double GetBalance() const { return balance_; }
+  inline Inventory GetInventory() const { return inventory_; }
```
CristianIonutIonescu, line 23: Return reference to avoid unnecessary copies

H5/include/store.h
```diff
@@ -0,0 +1,31 @@
...
+
+protected:
```
CristianIonutIonescu, line 25: Do you plan to derive this class if you set them protected?

H5/main.cc
```diff
@@ -0,0 +1,83 @@
...
+
+  addProds(store, apple, 2);
+  addProds(store, beer, 2);
+  addProds(store, chips, 3);
+  addProds(store, water, 5);
+
+  std::cout << "The store has: \n";
+  printer->print(store);
+
+  std::shared_ptr<Client> client = std::make_shared<RegularClient>("John", 17, 100);
```
CristianIonutIonescu, line 39: You don't handle your exception.Even though based on the data you sent it won't be thrown, you must handle your exceptions every time. Most likely this applies to your other methods as well

H5/include/Client/premium_client.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _GENERAL_STORE_PREMIUM_CLIENT_H_
+#define _GENERAL_STORE_PREMIUM_CLIENT_H_
+
+#include "client.h"
+
+class PremiumClient : public Client
+{
+public:
+  PremiumClient(std::string name, int age, double balance)
+    : Client(name, age, balance, 0.25) {}
```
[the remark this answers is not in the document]
aronkis: Yes. Why not?
Isn't premium just about the discount?

H5/include/store.h
aronkis: The idea behind this was that I might want to open a hardware store, with different properties so I made the variables protected to make the class future-proof.

H5/include/printer.h
```diff
@@ -1,27 +1,27 @@
 #ifndef _GENERAL_STORE_PRINTER_H_
 #define _GENERAL_STORE_PRINTER_H_
 
-#include "../include/Client/client.h"
-#include "../include/Product/product.h"
-#include "../include/store.h"
+class Store;
+class Product;
```
CristianIonutIonescu, line 5: unless I'm missing something you still don't use product here. It can be removed completely

H5/main.cc
```diff
@@ -10,74 +10,97 @@
 #include "include/Product/Drink/alcoholic.h"
 #include "include/Product/Drink/non_alcoholic.h"
 
-void addProds(std::shared_ptr<Store> store, std::shared_ptr<Product> product, int count)
+void addProds(Store* store, std::shared_ptr<Product> product, int count)
```
CristianIonutIonescu, line 13: input parameters should be sent by reference.

H5/main.cc
```diff
@@ -10,74 +10,97 @@
...
+    std::cerr << e.what() << '\n';
+  }
+  
+ 
+  Store store("ABC", 10);
+  
+  addProds(&store, apple, 2);
+  addProds(&store, beer, 2);
+  addProds(&store, chips, 3);
+  addProds(&store, water, 5);
```
CristianIonutIonescu, line 78: what happens if one of your product throws an exception while it's created? You handle the exception when you creat it but here you add it even if it might not have been created.

CristianIonutIonescu (approved):
I consider the homework done. There are a few things that have slipped but not critical. If you can update them until tomorrow's homework it's fine, if not please do it by next week homework (you'll have issues in the future with code logic if not).

CristianIonutIonescu:
For next homeworks, please don't add 2 different folders with the same classes for review. It was very hard for me to review it (you might have seen that I have commented to previous homework code as well)

marisvali (approved):

PR #35, opened 2023-03-07
Homework 4 lpitu

Lys4nder:


products.cpp
```diff
@@ -0,0 +1,93 @@
...
+beverages::beverages(std::string name, float price, int quantity) : product(name, price, quantity)
+{}
+
+nonAlcoholicDrinks::nonAlcoholicDrinks(std::string name, float price, int quantity) : beverages(name, price, quantity)
+{}
+
+alcoholicDrinks::alcoholicDrinks(std::string name, float price, int quantity) : beverages(name, price, quantity)
+{}
+
+bool nonAlcoholicDrinks::isSafe()
```
CristianIonutIonescu, line 39: nothing wrong with them but I encourage you to read a little bit about `constexpr`

customer.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+float customer::getMoneySpent()
+{
+    return moneySpent_;
+}
+
+void customer::spendMoney(float s)
+{
+    moneySpent_ += s;
+}
```
CristianIonutIonescu, line 24: nothing wrong but I suggest reading about inline methods

main.cpp
```diff
@@ -0,0 +1,47 @@
+#include <iostream>
+#include <vector>
+#include "products.hpp"
+#include "customer.hpp"
+
+int main()
+{
+    system("clear");
```
CristianIonutIonescu, line 8: indentation

CristianIonutIonescu (commented):
Great job with this draft. I've left some comments there but nothing that would require a `Request Change` review. 
PS: some comments that I have made regarding customers are true for products as well

customer.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+    return getAge() > 17;
+}
+
+float regularCustomer::getDiscount(float sum)
+{
+    //discount is in percentages
+    sum -= sum*discount_/100;
+    return sum;
+}
```
marisvali, line 45: This method doesn't get the discount, it applies the discount. When you have a getDiscount and a member called discount_ I immediately think getDiscount will return the value of discount_. The function name should save the reader this confusion. Maybe something like getPriceAfterDiscount or getDiscountedPrice.

products.hpp
```diff
@@ -0,0 +1,72 @@
...
+    private:
+        std::string name_;
+        float price_;
+        int quantity_;
+};
+
+class beverages : public product {
+    public:
+        beverages(std::string name, float price, int quantity);
+        virtual bool isSafe() = 0;
```
marisvali, line 23: It's fun to add new functions in the middle of the hierarchy, because it justifies the hierarchy. Only some products are beverages and only beverages are safe or not safe. Only some products are food and only foods can be eco or not. It makes sense so far.

However, now you have the following problem: the whole point of polymorphism is to use the objects of derived classes through their base class. So should the store contain a list of products? But then it can't find out if a product is safe. Then it needs to keep a list of foods and a list of beverages. But then whenever I want to do something with my whole inventory, I have to work with 2 lists instead of one. Example:

```
void clearInventory()
{
    foods.clear();
    beverages.clear();
}

int inventoryValue()
{
   int total = 0;
   for (auto food: foods)
   {
      total += food.getPrice();
   }
   for (auto beverage: beverages)
   {
      total += beverage.getPrice();
   }
}
```

Whenever a customer wants to buy a product, the store needs to check if the product is still available. Again, it needs to check which category the product belongs to then check the right list to see if the product is available. But this is exactly the problem that polymorphism tries to solve. And the point of the homework is to show how you can use polymorphism and SOLID to solve problems you would have without them.

marisvali (changes requested):
The project as it is doesn't use polymorphism and doesn't show SOLID principles in action.

"Extend(or implement from scratch) a general store"
"increasing income"
"Class for printing"
"customer enters shop, ask for a product"

products.cpp
Lys4nder: you mean to use constexpr for isSafe or constructor?

customer.cpp
```diff
@@ -0,0 +1,56 @@
+#include "customer.hpp"
+
+customer::customer(std::string name, int age) : name_(name), age_(age)
+{}
+
+inline const std::string& customer::getName()
+{
+    return name_;
+}
+
```
CristianIonutIonescu, line 10: Inline functions are usually defined inside class declaration (within the header files)

main.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <memory>
+#include "products.hpp"
+#include "customer.hpp"
+
+int main()
+{
+    system("clear");
+    
+
+    std::vector<vegetables> veggies;
```
CristianIonutIonescu, line 12: This whole block of code(reviewing from phone) is dangerous, you're assigning the addresses of these objects to another vector. If you were to limit the lifetime of your vectors, the one containing pointers to them would result in crash at best, indeterminate behavior.
You must dynamically allocate your memory for polymorphism

store.hpp
```diff
@@ -0,0 +1,26 @@
...
+        void setProducts(std::vector<product>);
+  
+    private:
+        std::string name_;
+        float moneyMade_;
+
+};
+
+
+class printStore {
```
CristianIonutIonescu, line 19: I know it's not final, but from the looks of it, you're not making use of polymorphism here. Printer shouldn't care what type of costumer or product it prints.

CristianIonutIonescu (commented):
Left a couple more comments. Please have a look

CristianIonutIonescu (approved):

marisvali (approved):

PR #36, opened 2023-03-08
draft added

914-Mereu-Bianca:


customer/Customer.cpp
```diff
@@ -0,0 +1,71 @@
+#include "Customer.h"
+
+Customer::Customer() : name(""), age(0), shopping_cart() {}
```
CristianIonutIonescu, line 3: I don't see a default constructor on the derived classes and as you have pure virtual methods in it you cannot instantiate this class, I recommend removing it as it serves no purpose

CristianIonutIonescu (commented):
Good job. Left some comments suggesting additional reading material. Looking forward to the final PR

store/Store.h
```diff
@@ -0,0 +1,36 @@
...
+        void setInventory(std::map<Product, int> _inventory);
+        void setIncome(int _income);
+
+        void purchase(Product prod, int quantity);
+
+        std::string toString();
+
+    private:
+        int income;
+        std::map<Product, int> inventory;
```
CristianIonutIonescu, line 32: read about https://www.learncpp.com/cpp-tutorial/object-slicing/ and how it can be prevented in your case

customer/Customer.h
```diff
@@ -0,0 +1,53 @@
...
+    public:
+        Customer();
+        Customer(std::string _name, int _age, std::map<Product, int> _shopping_cart);
+        Customer(const Customer& other);
+        Customer(Customer&& other);
+
+        Customer& operator=(const Customer& other);
+        Customer& operator=(Customer&& other);
+
+        ~Customer();
```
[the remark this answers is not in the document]
marisvali: To expand on what Cristi said: don't implement these unless needed.
https://en.cppreference.com/w/cpp/language/rule_of_three
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero
We are in the 'rule of zero' case.

This might be confusing as we asked you to implement these in the previous two homeworks. Homework 2 was a practice run specifically for implementing constructors/assignment/destructors. Homework 3 dealt with RAII and classes which are owners of raw pointers. So for these two assignments you had to implement these operations. From now on, implement only what is necessary.

Quote from the first link: "Classes that have custom destructors, copy/move constructors or copy/move assignment operators should deal exclusively with ownership (which follows from the Single Responsibility Principle). Other classes should not have custom destructors, copy/move constructors or copy/move assignment operators."

This applies to all classes in this assignment, follow the rule of zero if you can.

customer/Customer.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+RegularCustomer::RegularCustomer(std::string _name, int _age, std::map<Product, int> _shopping_cart)
+: Customer(_name, _age, _shopping_cart) {}
+
+double RegularCustomer::checkout() {
+    double total = 0;
+
+    for(std::map<Product, int>::iterator it = shopping_cart.begin(); it!=shopping_cart.end(); ++it){
+        total += it->first.getPrice() * it->second;
+    }
```
marisvali, line 55: This is not a mistake, just a syntactical sugar suggestion. I've seen you use ranged for loops before. You can do the same with map, to have a cleaner for:

```
for (auto& pair: shopping_cart) {
     total += pair.first.getPrice() * pair.second;
}
```

or, for more clarity:

```
for (auto& [product, quantity]: shopping_cart) {
    total += product.getPrice() * quantity;
}
```

customer/Customer.h
```diff
@@ -0,0 +1,53 @@
...
+
+        std::string getName();
+        virtual double checkout() = 0;
```
marisvali, line 21: I would use another name for the 'checkout' method. It sounds like when I call it I will "perform checkout". But its role is to compute the total cost of the items in the cart. Something like 'getCartValue' or 'getTotalCost' or 'computeTotalCost'.

marisvali (changes requested):
The project doesn't show polymorphism or SOLID in action. There aren't multiple products. Too much is missing to provide more feedback or guidance.

"increasing income"
"Multiple types of products"
"simulate multiple shopping situations"

A small note regarding the Repository class: it shouldn't know about a file path. The ReadWrite method should be concerned about the file path, and then ReadWriteFromURL can be concerned about an url, and ReadWriteFromDatabase can be concerned about a database connection string. The Repository's job is only as a container of Products. However, it isn't clear yet in this assignment what value it has over just using a 'std::vector<Product>' directly instead of a Repository.

store/Store.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if(this != &s){
+        income = std::move(s.income);
+        inventory = std::move(s.inventory);
+    }
+    return *this;
+}
+
+Store::~Store() {}
+
+std::map<Product, int> Store::getInventory()
```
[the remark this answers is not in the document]
914-Mereu-Bianca: here I should also return the inventory by reference

store/Store.cpp
```diff
@@ -0,0 +1,67 @@
...
+    income = _income;
+}
+
+std::string Store::toString()
+{
+    std::string total = "the prod in inventory are:\n";
+    for(auto elem : inventory){
+        total += std::to_string(elem.second) + " x " + elem.first.getName() + "\n";
+    }
+    return total;
```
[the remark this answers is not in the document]
914-Mereu-Bianca: I suppose it is about not writing it in the txt file, here I just wanted to test that everything is stored properly, but I will write the content to the txt file, sorry for any confusion

store/Store.cpp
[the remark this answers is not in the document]
CristianIonutIonescu: Correct. For your use case it would improve the performance. If you plan to use the inventory outside your store object lifetime, you should keep it as is.

store/Store.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    return inventory;
+}
+
+int Store::getIncome()
+{
+    return income;
+}
+
+void Store::setInventory(std::map<Product, int> _inventory)
```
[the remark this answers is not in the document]
CristianIonutIonescu: Corect, passing by value creates an unnecessary temporary object. For basic data types - int, char (not strings), float, double etc passing by value doesn't have a significant impact but passing objects does. Also a good practice is to make input parameters const

model/Product.h
```diff
@@ -8,18 +8,16 @@ class Product
...
-    //copy operator
-    Product& operator=(Product const &p); 
-    //move operator
-    Product& operator=(Product &&p) noexcept;
+    Product(std::string _name, int _price);
+    
+    Product(const Product & p) = default;
+    Product(Product&& p) noexcept = default; 
+    Product& operator=(Product const &p) = default;
+    Product& operator=(Product &&p) noexcept= default;
```
marisvali, line 17: You can delete these.

readWrite/ReadWrite.cpp
```diff
@@ -1,25 +1,14 @@
...
 
-void ReadWrite::write()
+void ReadWrite::write(Store& store)
 {
-    std::ofstream fout(repository.getFile());
-    for(auto product : repository.getProducts()) {
-        fout << product << '\n';
-    }
+    std::ofstream fout(file);
+    fout << store << '\n';
```
marisvali, line 11: The logic for writing the store from a file is inside the Store class now. The ReadWrite class doesn't add any value in the project. Might as well do:
```
std::ofstream fout(file);
fout << store << '\n';
```
instead of
```
ReadWrite rw( _file);
rw.write(store);
```
The homework assignment asked for "Class for printing store inventory/premium customer to console/file(should be extendable)". ReadWrite doesn't satisfy the requirement.

store/Store.h
```diff
@@ -3,33 +3,37 @@
...
-        void purchase(Product prod, int quantity);
+        bool operator==(Store const& s) const;
 
+        friend std::ostream& operator<<(std::ostream& writer, const Store &store);
         std::string toString();
 
     private:
         int income;
-        std::map<Product, int> inventory; 
+        std::map<std::shared_ptr<Product>, int> inventory;
```
marisvali, line 36: I think I understand why you want to use something like map<Product, Quantity>. You want to quickly look up if a product exists. It's efficient and elegant. BUT! You will always have a problem when you try to do this using a base class as the key. That's because the key isn't a Product object. It's a Product*, which is a pointer, which has the value of an address. So the map isn't checking if two Product objects are the same, it's checking if two pointers are the same. Which means when I run this code:
```
    std::map<std::shared_ptr<Product>, int> inventory;
    std::shared_ptr<Product> p1 = std::make_shared<Soda>("orange juce", 15);
    std::shared_ptr<Product> p2 = std::make_shared<Alchool>("gin", 120);
    inventory[p1] = 12;
    inventory[p2] = 30;
    Store store = Store(0, inventory);
    std::shared_ptr<Product> p3 = std::make_shared<Soda>("orange juce", 15);
    std::cout << store.existProduct(p1, 1) << " " << store.existProduct(p3, 1) << "\n";
```
I get "1 0" as output instead of "1 1". Does all that make sense? This will happen with shared_ptr and unique_ptr, there's no getting away from it.

marisvali (changes requested):
This is still missing:

"Class for printing store inventory/premium customer to console/file(should be extendable)"

Trying to use a base class as key in a map is going to cause nothing but trouble. The fix isn't trivial either, you have to rethink some things. Also, you should always use unordered_map unless you have a good reason to use map.

store/Store.h
914-Mereu-Bianca: I understand, I was thinking of a way to make it work and I come up with an idea. So I could add a function getProduct in the Store as it follows: 
`std::shared_ptr<Product> const Store::getProduct(const std::shared_ptr<Product>& prod, int quantity)
{
    for(auto& [first, second] : this->inventory){
        if(*first == *prod) 
            if(second < quantity)
                return nullptr;
            else return first;
    }
    return nullptr;
}`
and whenever the function buyProduct is called I will check through this function if the product exists and the rest of the code will continue with the correct address of the product that is in the map. 
Is this okay or should I give up on using the map?

PR #37, opened 2023-03-08
Upload homework 4

oprisor-adrian:


course_4_homework/include/customer/customer.h
```diff
@@ -0,0 +1,37 @@
...
+    Customer();
+    Customer(const std::string&, const std::string&, 
+	     const std::map<Product, uint8_t>& = {});
+    Customer(const Customer&);
+    Customer(Customer&&);
+    Customer& operator=(const Customer&);
+    Customer& operator=(Customer&&);
+    ~Customer();
+    // return the amount that customer have to pay
+    virtual double CheckOut(Store&) const;
```
marisvali, line 21: CheckOut sounds like it says "perform checkout". But the function actually returns the amount that the customer has to pay. You added the comment, which means the name of the function doesn't properly describe what the function does. Try to find a better name.

course_4_homework/include/product.h
```diff
@@ -0,0 +1,29 @@
...
+    void SetAgeLimit(const uint8_t&);
+    std::string GetName() const;
+    std::string GetCategory() const;
+    double GetPrice() const;
+    uint8_t GetAgeLimit() const;
+  private:
+    std::string name_;
+    std::string category_;
+    double price_;
+    uint8_t age_limit_;
```
marisvali, line 27: Careful. You've defined a very flexible Product class which means there's need or room for a hierarchy of different kinds of products. This works well if you just want to have a store that works. But the point of the homework is to practice polymorphism and SOLID. I suggest making the Product class less flexible and creating a hierarchy of products.

course_4_homework/src/customer/customer.cc
```diff
@@ -0,0 +1,106 @@
...
+  for(const auto& [product, quantity] : shopping_cart_){
+    amount += product.GetPrice() * quantity;
+    market.SellProduct(product, quantity);
+  }
+  market.AddIncome(amount);
+  return amount; 
+}
+
+void Customer::AddToCart(const std::pair<Product, uint8_t>& item){
+  if(item.first){
```
marisvali, line 55: The if is redundant. This makes sense in Java/C# where all objects are references and they can be null. However in C++ objects are values and item.first will always be of type Product, which will always be non-null (only Product* can be null).

course_4_homework/src/customer/customer.cc
```diff
@@ -0,0 +1,106 @@
...
+void Customer::RemoveProduct(const Product& product, const uint8_t& amount){
+  if(auto search = shopping_cart_.find(product); search != shopping_cart_.end()){
+    if(amount == 0 || search->second - amount < 0){
+      shopping_cart_.erase(product);
+      return;
+    }
+    shopping_cart_[product] = search->second - amount;
+    return;
+  } 
+  std::cout << "Product " << product.GetName() << " no't found!\n";
```
marisvali, line 81: Don't write more than one instruction in a single line. C++ is very flexible and lets you do all sorts of things. But you should always put separate instructions on separate lines. Instead of
`if(auto search = shopping_cart_.find(product); search != shopping_cart_.end()){`
do
```
auto search = shopping_cart_.find(product);
if(search != shopping_cart_.end()){
```
Spelling error in "no't found".
And if you're using early returns, you can do this:
auto search = shopping_cart_.find(product);
```
  if(search == shopping_cart_.end()){
    std::cout << "Product " << product.GetName() << " no't found!\n";
    return;
  }
  if(amount == 0 || search->second - amount < 0){
    shopping_cart_.erase(product);
    return;
  }
  shopping_cart_[product] = search->second - amount;
```

marisvali (changes requested):
The project doesn't show polymorphism or SOLID in action. There aren't multiple products. Too much is missing to provide more feedback or guidance.

"Multiple types of products"
"Class for printing"
"In main, simulate multiple shopping situations"

course_4_homework/include/printer.h
```diff
@@ -0,0 +1,35 @@
...
+  public:
+    Printer(const object obj);
+    void Print();
+    void Write(const std::string& filename);
+  private:
+    object obj_;
+};
+#endif  // STORE_SIMULATOR_INCLUDE_PRINTER_H_
+
+template<class object>
```
CristianIonutIonescu, line 18: Although it works this way, the scope of the requirement was to use polymorphism here.

course_4_homework/include/printer.h
```diff
@@ -0,0 +1,35 @@
...
+class Printer{
+  public:
+    Printer(const object obj);
+    void Print();
+    void Write(const std::string& filename);
+  private:
+    object obj_;
+};
+#endif  // STORE_SIMULATOR_INCLUDE_PRINTER_H_
```
CristianIonutIonescu, line 16: If you add your endif here, all below it will be included in each file you include this header. #endif should be at the end of the file

CristianIonutIonescu (approved):
Although there are many things that can be addressed, I consider that the majority of the requirements were met on a level enough for me to consider the homework done.

course_4_homework/main.cc
```diff
@@ -0,0 +1,38 @@
...
+#include "include/customer/customer.h"
+#include "include/customer/underage_customer.h"
+
+int main(){
+  std::map<Product*, uint32_t> items;
+  Product* prod1 = new Book("BookName", 25.6, "BookAuthor", 145);
+  items.insert({prod1, 20});
+  Product* prod2 = new Toy("Toy",45.1234, 15);
+  items.insert({prod2, 30});
+  Store store(0.0, "MyStore", "somewhere", items);
```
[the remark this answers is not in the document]
oprisor-adrian: I made a little mistake here. I did some tests using smart pointers and forgot to change them. Thank you for pointing out the mistake.

PR #38, opened 2023-03-08
Tema 4: extension from Tema2

radavid2017:


tema2DHoka/file.h
```diff
@@ -15,4 +16,5 @@ class StoreIO {
     void write(Store& store);
     private:
     std::string name;
```
marisvali, line 18: Since file.h/file.cpp will contain the StoreIO class, rename the files to storeio.h/storeio.cpp.
I have a curiosity regarding file.cpp so I'll leave it here. You have this code:
```
if (!store.isEmpty()) {
for (auto itr = store.getData().begin(); itr < store.getData().end(); ++itr) {
    my_file << itr->getNume() << " " << itr->getCantitate() << " " << itr->getPret() << std::endl;
}
// for (Product& product : store.getData()) {
//     my_file << product.getNume() << " " << product.getCantitate() << " " << product.getPret() << endl;
// }
```
How come you opted for the version with the iterator instead of the ranged for loop? It's not a mistake, I'm just curious.

tema2DHoka/main.cpp
```diff
@@ -1,17 +1,35 @@
...
+    // storeIO.write(store);
+    // store.displayProduct("apa");
+    // Product a = Product("ceva",1,2);
+    // Product b = Product(a);
+    // std::cout<<b;
+
+    Customer* underAge = new UnderAge(5);
+    Customer* adult = new Adult(34);
+    Product* p1 = new Alcohol("Vodka");
+    Product* p2 = new Feed("Paine");
```
marisvali, line 25: You should never use raw pointers like this, if you can avoid it. Use smart pointers, unless you have a good reason not to.
`std::unique_ptr<Customer> underAge = std::make_unique<Customer>();`
Your code as it is has memory leaks (you never call delete). In this case it doesn't affect the execution, the application ends immediately and the OS will free the memory. But the cases where memory leaks are acceptable are rare and decided by experts. For now, consider all memory leaks forbidden.

tema2DHoka/product.cpp
```diff
@@ -11,6 +11,7 @@ Product::Product(const Product& other) {
     nume = other.nume;
     cant = other.cant;
     pret = other.pret;
+    age_restricted = other.age_restricted;
```
marisvali, line 14: The field is initialized here, but it's not initialized in the default constructor. The derived classes initialize it, yes. But someone who derives from the class might not remember to do that. You are relying on the fact that another programmer will know the convention is that any derived class must initialize this member. Don't rely on conventions when you can enforce something. It's very dangerous to leave any variable uninitialized. It will get a random value. Maybe with one value the software will work, maybe with another, it won't. This is the most expensive bug: the one that happens only once in a while for no apparent reason. Never leave any variable uninitialized.

tema2DHoka/product.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef _PRODUCT_
+#define _PRODUCT_
 #include <iostream>
 
 class Product {
```
marisvali, line 5: The Product class has no pure virtual function, so I can create this:
Product product("etc", 3, 2);
Which might not be what you intended. You have two derived products: Alcohol and Feed. One is age restricted, one isn't. They don't differ from Product in any other way, so they cover all possible products. In fact, a plain Product object will have its age_restricted member uninitialized.

tema2DHoka/store.cpp
```diff
@@ -13,20 +13,27 @@ Store::Store(std::vector<Product> data_) {
     std::cout << "S-au construit datele unui store\n";
 }
 
+Store::Store(std::vector<Product> data_store, int income_store) : data(data_store), income(income_store) {
```
marisvali, line 16: Make a habit to pass all containers as const&, otherwise you're making unnecessary copies.

tema2DHoka/underage.cpp
```diff
@@ -0,0 +1,43 @@
...
+    switch (askForProduct(name, store))
+    {
+    case -1:
+        std::cout<<"Age restricted\n";
+        break;
+    
+    default:
+        Customer::buy(name, store);
+        break;
+    }
```
marisvali, line 42: If you have only two options, use if instead of switch.

tema2DHoka/underage.cpp
```diff
@@ -0,0 +1,43 @@
...
+            
+            return -1;
+        }
+        else {
+            
+            return 1;
+        }
+    }
+    else {
+        return 0;
```
marisvali, line 28: Use enums instead of integers. Google around to see why this is a good idea and how to do it in C++.

tema2DHoka/underage.h
```diff
@@ -0,0 +1,15 @@
...
+#define _UNDER_AGE_
+
+#include "customer.h"
+
+class UnderAge : public Customer {
+    public:
+        UnderAge();
+        UnderAge(int age_for_under_age);
+        ~UnderAge();
+        int askForProduct(std::string const& name, Store& store) const noexcept;
```
marisvali, line 11: Use the override keyword on functions that override virtual functions. C++ allows us not to do it because of backwards compatibility, but everyone should use the override keyword in these cases.

tema2DHoka/main.cpp
```diff
@@ -1,17 +1,35 @@
...
-    storeIO.read(store);
-    std::cout << store;
-    store.delProduct("paine");
-    std::cout << store;
-    storeIO.write(store);
-    store.displayProduct("apa");
-    Product a = Product("ceva",1,2);
-    Product b = Product(a);
-    std::cout<<b;
+    store.addProduct(*p1);
```
marisvali, line 27: You're slicing your object here (https://www.learncpp.com/cpp-tutorial/object-slicing/). The whole point of polymorphism is for the store to hold a collection of pointers to Product objects, and each object can be any derived type. The store will treat them all the same, but they will act different when they need to. If you hold a collection of values, not pointers, you don't have polymorphism anymore.

marisvali (changes requested):
The project follows the requirements fairly well. It uses polymorphism and SOLID (the derived customers don't rely on the derived products directly, they are connected through their base interfaces/classes). However, it's minimal. A few more features would be nice.

tema2DHoka/file.h
radavid2017: Honestly, my first try was with the iterator. Before adding the commented code which is the second option, my first way had some error and somehow after a while of code editing the loop with the iterator worked well. So I chose to let with iterator because it seems more professionally and maybe is efficient with the memory, but I am not sure yet.

tema2DHoka/main.cpp
radavid2017: I will take it under consideration, thank you for giving the information. I was wondering why does not appear the destructor message in the console, but now it make sense.

tema2DHoka/product.cpp
radavid2017: I took that under consideration and have just modified the constructors to work properly in any situation. Thank you for making me understand its initialization importance.

tema2DHoka/product.h
radavid2017: Yes, I was wondering if just initializing some methods with "virtual" would make the class abstract. Now I see the main feature of an abstract class. I chose to define the destructors of the Product and Customer as pure virtual functions. I hope it is okay. If not, please let me know.

tema2DHoka/store.cpp
radavid2017: Thank you for telling me about it. I will look to make it a habit.

tema2DHoka/adult.cpp
```diff
@@ -0,0 +1,13 @@
+#include "adult.h"
+
+Adult::Adult() {
+    std::cout<<"Adult constructor apelat\n";
+}
+
+Adult::Adult(int age_adult) : Customer(age_adult) {
+    std::cout<<"Adult construt apelat\n";
+}
```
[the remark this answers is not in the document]
radavid2017: I added restrictions to underage and adult constructors. The method named isAdult() is removed to avoid conflicts and ambiguity.

tema2DHoka/alcohol.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+Alcohol::Alcohol(std::string const name, int pret) : Product(name, 10, pret) {
+    setAgeRestricted(true);
+    std::cout<<"Alcohol constructor apelat\n";
+}
+
+Alcohol::Alcohol(std::string const name, int cant, int pret) : Product(name, cant, pret) {
+    setAgeRestricted(true);
+    std::cout<<"Alcohol constructor apelat\n";
+}
```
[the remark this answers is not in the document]
radavid2017: I will take it under consideration and from now on I will write the code just in english.

tema2DHoka/customer.h
```diff
@@ -0,0 +1,28 @@
+#ifndef _CUSTOMER_
+#define _CUSTOMER_
+
+#include <iostream>
+#include <string>
+#include "product.h"
+#include "store.h"
+
+class Customer {
```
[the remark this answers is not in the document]
radavid2017: I made the destructors of bases as pure virtual functions. I hope it was a good idea to make it abstract class. If not, let me know.

tema2DHoka/main.cpp
radavid2017: I worked this time with smart pointers, and it seems to work just as it should be now.

tema2DHoka/customer.h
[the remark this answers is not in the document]
CristianIonutIonescu: Although you can make destructors pure virtual, you'd still need to declare a function body for them, which somehow defeats the purpose of pure virtual functions. I'd keep them just virtual and that's it.

tema2DHoka/store.h
```diff
@@ -1,27 +1,38 @@
...
-    bool displayProduct(std::string searchName);
+    bool displayProduct(std::string& searchName);
     void display();
+    bool sellProduct(std::string const& name);
+    void increaseIncome(int const pret);
     private:
-    std::vector<Product> data;
-};
\ No newline at end of file
+    std::vector<Product*> data;
```
marisvali, line 34: At first I thought this as an error, but now I see that there's no memory leak. The store just holds references to Products, and the owner of the Product objects is the main function. But it would make more sense for the store to own the products, so a vector of unique_ptr.

tema2DHoka/store.h
```diff
@@ -1,27 +1,38 @@
...
-    Store(std::vector<Product> data_);
+    Store(std::vector<Product*>& data_);
+    Store(std::vector<Product*>& data_store, int income_store);
     Store (const Store &other);
     Store& operator=(Store &other);
     Store() {}
     // MOVE CONSTRUCT
     Store(Store&& other) noexcept;
     Store* operator=(Store&& other) noexcept;
     ~Store();
```
marisvali, line 19: You don't need any of these.

tema2DHoka/storeio.cpp
```diff
@@ -1,10 +1,12 @@
...
+#include "alcohol.h"
+#include "food.h"
 #include <string>
 #include <regex>
 
 StoreIO::StoreIO(std::string file_name) {
     name = file_name;
-    std::cout << "StoreIO construct este apelat\n";
+    std::cout << "StoreIO constructor\n";
 }
```
marisvali, line 10: The StoreIO class isn't showcased in main. It would be nice to see it used.

marisvali (approved):
This implements enough of the requirements to qualify as done. There are tweaks that can be made here and there, but this homework already took longer to implement than we estimated, so we might as well stop here for now.

CristianIonutIonescu (approved):
As Vali said, even though there are many things that could be improved the curent status of the PR is enough for me to consider the homework done

tema2DHoka/customer.cpp
```diff
@@ -0,0 +1,34 @@
...
+const product_situation Customer::askForProduct(std::string const& name, Store const& store) const noexcept {
+    return store.isAlready(name) ? valid_for_buy : unfound_product;
+}
+
+void Customer::buy(std::string const& name, Store& store) {
+    if (askForProduct(name, store) == valid_for_buy) {
+        store.sellProduct(name);
+    }
+    else {
+        std::cout<<"Produsul unfound\n";
```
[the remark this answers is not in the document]
radavid2017: I am sorry, it seems some words escaped from editing. Maybe it is because of a little fatigue.

tema2DHoka/main.cpp
```diff
@@ -1,17 +1,40 @@
...
+#include "customer.h"
+#include "underage.h"
+#include "adult.h"
+#include "alcohol.h"
+#include "food.h"
+#include "writeTxt.h"
+#include <memory>
 
 int main() {
-    StoreIO storeIO = StoreIO("/home/yuda/2022/tema2DHoka/date.txt");
```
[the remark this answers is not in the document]
radavid2017: I had some problems in the previous homework with this. I was giving the relative path and the file was not updated. I will try to use the relative path from now on, I hope it will work these times.

tema2DHoka/store.h
radavid2017: This was a question which dropped out of my mind. I was thinking about it but I forgot to ask. Thank you for letting me know this information!

tema2DHoka/storeio.cpp
radavid2017: It was used in the second homework, but it works very similar with the Write system that has been implemented now.

PR #41, opened 2023-03-12
HW5 lpitu

Lys4nder:
changed names according to naming convention


store.cpp
```diff
@@ -0,0 +1,87 @@
+#include "store.hpp"
+
+Store::Store(std::string name) : name_(name)
+{}
+
+void Store::AddCustomer(std::unique_ptr<Customer> cust)
+{
+    customers_.emplace_back(std::move(cust));
```
CristianIonutIonescu, line 8: As Vali explained in the previous lesson std:: move does nothing

customer.hpp
```diff
@@ -0,0 +1,33 @@
...
+    private:
+        std::string name_;
+        int age_;
+        float moneySpent_ = 0;
+};
+
+class RegularCustomer : public Customer {
+    public:
+        RegularCustomer(std::string name, int age);
+        float GetDiscount(float sum);
```
CristianIonutIonescu, line 21: Mark it as override

customer.hpp
```diff
@@ -0,0 +1,33 @@
...
+    private:
+        static constexpr int kDiscount_ = 5;
+};
+
+class PremiumCustomer : public Customer {
+    public:
+        PremiumCustomer(std::string name, int age);
+        float GetDiscount(float sum);
```
CristianIonutIonescu, line 29: Mark it as override

customer.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+class Customer {
+    public:
+        Customer(std::string name, int age);
+        inline bool IsAdult() { return age_> 17; }
+        inline const std::string& GetName() { return name_; }
+        inline int GetAge() { return age_; }
+        inline float GetMoneySpent() { return moneySpent_; }
+        inline void SpendMoney(float sum) { moneySpent_+= sum - GetDiscount();}
+        inline virtual float GetDiscount() { return 0;}
```
CristianIonutIonescu, line 11: Don't inline virtual functions

customer.hpp
```diff
@@ -0,0 +1,33 @@
+#include <iostream>
```
CristianIonutIonescu, line 1: Don't include headers that are not needed in header files. Include them where you use them i.e source files

products.hpp
```diff
@@ -0,0 +1,74 @@
+#include <iostream>
```
CristianIonutIonescu, line 1: Don't include headers that you don't use in header files. Include them in source files

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <stdlib.h>
+#include <time.h>
+
+class Product {
+    public:
+        Product(std::string name, float price, int quantity);
+        inline const std::string& GetName() { return name_; }
+        inline int GetQuantity() { return quantity_; }
+        inline float GetPrice() { return price_; }
+        inline virtual void PrintNameQuantity() {std::cout << "You have " << GetQuantity() << " of " << GetName() << '\n';}
```
CristianIonutIonescu, line 11: Don't inline virtual functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+    private:
+        std::string name_;
+        float price_;
+        int quantity_;
+};
+
+class Beverages : public Product {
+    public:
+        Beverages(std::string name, float price, int quantity);
+        inline virtual bool IsSafe() = 0;
```
CristianIonutIonescu, line 25: Don't inline virtual functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+};
+
+
+class NonAlcoholicDrinks : public Beverages {
+    public:
+        NonAlcoholicDrinks(std::string name, float price, int quantity);
+        inline bool IsSafe() { return true; }
```
CristianIonutIonescu, line 32: Don't inline virtual functions, mark overriden functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+};
+
+class AlcoholicDrinks : public Beverages {
+    public:
+        AlcoholicDrinks(std::string name, float price, int quantity);
+        inline bool IsSafe() { return false; }
```
CristianIonutIonescu, line 38: Don't inline virtual functions, mark overriden functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+class Whiskey : public AlcoholicDrinks {
+    public:
+        Whiskey(std::string name, float price, int quantity);
+};
+
+class Food : public Product {
+    public:
+        Food(std::string name, float price, int quantity);
+        inline virtual bool IsEco() = 0;
```
CristianIonutIonescu, line 54: Don't inline virtual functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+        virtual void SetEco() = 0;
```
CristianIonutIonescu, line 55: Don't inline virtual functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+};
+
+class Vegetables : public Food {
+    public:
+        Vegetables(std::string name, float price, int quantity);
+        inline bool IsEco() { return eco_; }
```
CristianIonutIonescu, line 61: Don't inline virtual functions, mark overriden functions

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+        void SetEco();
```
CristianIonutIonescu, line 62: Mark it as override

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+    private:
+        bool eco_;
+};
+
+class Pastry : public Food {
+    public:
+        Pastry(std::string name, float price, int quantity);
+        bool IsEco() { return eco_; }
```
CristianIonutIonescu, line 70: Mark it as override

products.hpp
```diff
@@ -0,0 +1,74 @@
...
+        void SetEco();
```
CristianIonutIonescu, line 71: Mark it as override

store.cpp
```diff
@@ -0,0 +1,87 @@
...
+{}
+
+void Store::AddCustomer(std::unique_ptr<Customer> cust)
+{
+    customers_.emplace_back(std::move(cust));   
+}
+
+void Store::AddProduct(std::unique_ptr<Product> prod)
+{
+    products_.emplace_back(std::move(prod));
```
CristianIonutIonescu, line 13: As Vali explained in the previous lesson std:: move does nothing

CristianIonutIonescu (commented):
I have just reviewed the code from coding standard point of view. I might have missed some mistakes as I reviewed from phone but should give you an overview of what you need to improve

store.cpp
Lys4nder: std::move does something i believe, because Store class has a private member of unique pointer and I am passing the ownership when creating the products vector

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+void Store::RegisterMoney(float s)
+{
+    if (s < 0)
+        throw 2;
```
CristianIonutIonescu, line 19: Don't throw random numbers as exception.

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+            return;
+        }
+    }
+    try {
+        cust->SpendMoney(prod->GetPrice() * quantity);
+        RegisterMoney(prod->GetPrice() * quantity);
+        prod->SellQuantity(quantity);
+        std::cout << "Sold " << quantity << " " << productName << " to " << customerName << " for " << prod->GetPrice() * quantity << '\n';
+    }
+    catch (int & e) {
```
CristianIonutIonescu, line 121: Not a good pattern to handle exceptions. Create custom exceptions inheriting from std::runtime_error and handle each of it in a separate catch

store.hpp
```diff
@@ -0,0 +1,33 @@
+#include <iostream>
```
CristianIonutIonescu, line 1: Don't include what you don't use. Include it in source file.
Also, you can forward declare the client and product and include the header in source file.

CristianIonutIonescu (changes requested):
Main problem with the current solution is throwing ints and catching them.

store.cpp
CristianIonutIonescu: yes, you are right, you need the std::move here

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+void PrintStore::PrintCustomers() const 
+{
+    std::cout << "Customers for Store " << store_.GetName() << ":\n";
+    try {
+        for (const auto& cust : store_.GetCustomers())
+        {
+            if (cust->GetName().empty())
```
CristianIonutIonescu, line 48: instead of allowing a customer to have an empty name and throwing here, you'd better not allow creating a customer without name. Same for product

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+    catch (int & e) {
+        std::cout << "Exception caught: " << e << '\n';
+        if (e == 1)
+        {
+            std::cout << "Card declined, customer " << cust->GetName() << " does not have enough money\n";
+        }
+
+        if (e == 2) //rollback changes done
+        {
+            cust->SpendMoney(prod->GetPrice() * quantity * (-1)); //give money back to customer
```
CristianIonutIonescu, line 130: You are breaking S from SOLID here:
`every class, module, or function in a program should have one responsibility/purpose in a program.`
Write proper methods for returning the money.

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+    try {
+        for (const auto& cust : store_.GetCustomers())
+        {
+            if (cust->GetName().empty())
+                throw 1;
+                
+            std::cout << "Customer: " << cust->GetName() << " with " << cust->GetMoneySpent() << " spent money.\n";
+        }
+    }
+    catch(int &e) {
```
CristianIonutIonescu, line 54: not a valid case to throw and catch

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+            {
+                cust->SetName("Unknown");
+                break;
+            }
+    }
+}
+
+
+
+void Store::SellProduct(const std::string& productName, int quantity, const std::string& customerName)
```
CristianIonutIonescu, line 68: How do you inform your caller that the client bought something and give him the chance to recover from a failure? By caller I mean the code where you call this method. It cannot read the console output

CristianIonutIonescu (changes requested):
adding additional issues

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+    moneyMade_ += s;
+}
+
+void PrintStore::PrintInventory() const
+{
+    std::cout << "Inventory for Store " << store_.GetName() << ":\n";
+    try {
+        for (const auto& inv : store_.GetProducts())
+        {
+            if (inv->GetName().empty())
```
CristianIonutIonescu, line 29: you are dereferencing a pointer without checking if it's null. Same for products

CristianIonutIonescu (changes requested):
adding one more comment

store.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    customers_.emplace_back(std::move(cust));
+}
+
+void Store::AddProduct(std::unique_ptr<Product> prod)
+{
+    products_.emplace_back(std::move(prod));
+}
+
+void Store::RegisterMoney(float s)
```
CristianIonutIonescu, line 16: find a better name for the argument instead of `s`

store.hpp
```diff
@@ -0,0 +1,33 @@
...
+        void RegisterMoney(float s);
+    private:
+        std::string name_;
+        float moneyMade_ = 0;
+        std::vector<std::unique_ptr<Product> > products_;
+        std::vector<std::unique_ptr<Customer> > customers_;
+};
+
+
+class PrintStore {
```
CristianIonutIonescu, line 24: Don't name classes after an action but after an object.
PrintStore -> StorePrinter

store.cpp
Lys4nder: let me know if the current implementation is correct

CristianIonutIonescu: the caller (your main function) still doesn't know if the transaction was successful or not

store.cpp
```diff
@@ -0,0 +1,135 @@
...
+        if (transactionSucceeded == 3)
+        {
+            cust->RollbackSpendMoney(prod->GetPrice() * quantity);
+        }
+        if (transactionSucceeded == 4)
+        {
+            cust->RollbackSpendMoney(prod->GetPrice() * quantity);
+            RollbackRegisterMoney (prod->GetPrice() * quantity);
+        }
+    }
```
CristianIonutIonescu, line 133: these can be done in the catch bodies of your exceptions, no need to handle them separately

customer.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    try {
+        float s = sum - GetPriceAfterDiscount(sum);
+        moneySpent_+= s;
+        availableMoney -= s;
+    }
+    catch(const GetPriceAfterDiscountException& e)
+    {
+        std::cerr << "Exception caught: " << e.what();
+    }
```
CristianIonutIonescu, line 31: if you're catching here, the caller doesn't know if the money was spent or not. You can catch this exception upstream of this method (where it's called)

CristianIonutIonescu (approved):
Minor changes that need to be done. Approving it but make sure you fix them by next homework if you cannot do it by tomorrow morning.

store.cpp
```diff
@@ -13,58 +15,42 @@ void Store::AddProduct(std::unique_ptr<Product> prod)
...
-            {
-                cust->SetName("Unknown");
-                break;
-            }
+    if (store_.GetCustomers().empty())
+        return;
+    
+    for (const auto& cust : store_.GetCustomers())
+    {
+        std::cout << "Customer: " << cust->GetName() << " with " << cust->GetMoneySpent() << " spent money.\n";
```
CristianIonutIonescu, line 50: same here regarding pointers

PR #43, opened 2023-03-13
Upload homework 5

oprisor-adrian:
To do: 
- [x] Code styling
- [x] Use a file handler in FilePrinter class ( FILE* )
- [x] Logging class
- [x] **Error handling**

⚠️ In this branch is also pushed homework 4 (unmodified).


course_5_homework/include/customer/customer.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "../product/product.h"
+#include "../printer/printable.h"
+#include "../store.h"
+
+class Customer: virtual public Printable{
+  public:
+    Customer(const std::string& name, const std::string& address);
+    // return the amount that customer have to pay
+    virtual double CheckOut(Store& store);
```
enachebogdan, line 18: I'm curious why you need to return the amount at the end of the checkout.

course_5_homework/src/printer/file_handler.cc
```diff
@@ -0,0 +1,51 @@
...
+}
+
+int FileHandler::FileClose() {
+  int output = std::fclose(stream_);
+  stream_ = nullptr;
+  return output;
+}
+
+std::size_t FileHandler::FileRead(char* buffer,std::size_t counter){
+  return std::fread(buffer, sizeof(char), counter, stream_);
```
enachebogdan, line 46: Do you need to check if the file is open for read or write before?

course_5_homework/include/store.h
```diff
@@ -0,0 +1,71 @@
...
+#include <ostream>
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "./printer/printable.h"
+#include "./product/product.h"
+
+enum Status{
```
enachebogdan, line 13: have a look into enum class as well, and why is it better than plain enum.

course_5_homework/src/store.cc
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+  return Status::NotFound;
+} 
+
+bool Store::IsEmpty() const{
+  return store_name_.empty() && location_.empty() && income_ == 0.0 && inventory_.empty();
+}
+
+std::tuple<bool, std::unique_ptr<Product>> Store::Exist(const std::string& id) const{
```
enachebogdan, line 77: Same discussion for Exist, not sure what you are trying to accomplish, maybe you could write this in a clearer way.

enachebogdan (changes requested):
Good stuff Adrian, PR looks good in general, I expect a bit more on the exception handling part.
It's admirable that you try using multiple c++ features, discussed or not during the presentations, but make sure you don't over complicate things when it's not unnecessary.

course_5_homework/main.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "./include/printer/printer.h"
+#include "./include/printer/file_printer.h"
+#include "./include/printer/console_printer.h"
+#include "./include/printer/file_printer.h"
+#include "./include/product/product.h"
+#include "./include/product/toy.h"
+#include "./include/customer/premium_customer.h"
+
+#include "./include/logger.h"
+FilePrinter Logger::log_("course_5_homework/data/log.txt");
```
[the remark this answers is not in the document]
oprisor-adrian: Here _log__ is also a static data member and can be initialized outside the class using the scope resolution operator. I did this because I want the _log__ to be initialized only once in the entire program and then used (calling the _Log_ method, also static) wherever it is needed just by including the header. I can't access the _log__ from an instance of Logger object.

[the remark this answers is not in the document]
enachebogdan: The static part is handled right, I understand what you're trying to do and it's perfectly fine from that perspective, I was just curious about the "private" access modifier of log_. I would expect a compile error when you try to access from outside the class, so I was going to suggest maybe another static function to initialize it.

course_5_homework/include/customer/customer.h
```diff
@@ -0,0 +1,57 @@
...
+#include <cstdint>
+#include <memory>
+#include <unordered_map>
+#include <string>
+
+#include "../product/product.h"
+#include "../printer/printable.h"
+#include "../store.h"
+
+class Customer: virtual public Printable{
```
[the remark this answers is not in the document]
enachebogdan: I did suggest in my presentation that you look up virtual inheritance, and I'm glad you did. I thought maybe I was missing the "diamond" inheritance problem here

course_5_homework/include/printer/file_handler.h
```diff
@@ -0,0 +1,32 @@
...
+    std::size_t FileWrite(const char* buffer, std::size_t count);
+    void SetStream(std::FILE* stream);
+    std::FILE* GetStream() const;
+  private:
+    std::FILE* stream_;
+};
+inline void FileHandler::SetStream(std::FILE* stream){
+  stream_ = stream;
+}
+inline std::FILE* FileHandler::GetStream() const {
```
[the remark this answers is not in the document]
oprisor-adrian: In _Logger_ class i access the _std::FILE*_ through _FilePrinter_ class to allow me to access the _FileWrite_ method directly, and that method could not be constant. 
I don't know why I chose to do in this way. I'll remake it and remove/modify _GetStream()_ method. ^_^

course_5_homework/src/printer/file_printer.cc
```diff
@@ -0,0 +1,31 @@
...
+  if(file_.IsOpen()){
+    try{
+      std::ostringstream out;
+      printable.Print(out);  // can throw an exception
+      out << "\n";
+      std::string temp = out.str();
+      const char* buffer = temp.c_str();
+      auto buffer_size = file_.FileWrite(buffer, std::strlen(buffer));
+      if(buffer_size != std::strlen(buffer)){
+        throw std::runtime_error("error: buffer_size != std::strlen(buffer) when printing in file.");
```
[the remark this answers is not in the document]
oprisor-adrian: The error had to be caught in _main.cc_. I'll add more examples in _main_ to show that.

course_5_homework/src/printer/file_handler.cc
oprisor-adrian: Yes! I'll add a check before trying to read from file.

course_5_homework/src/customer/customer.cc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+bool Customer::IsEmpty() const{
+  return name_.empty() && address_.empty() && shopping_cart_.empty();
+}
+
+std::tuple<bool, std::unique_ptr<Product>> Customer::Exist(const std::string& id){
+  auto search = shopping_cart_.find(id);
+  if(search != shopping_cart_.end()){
+    return {true, std::make_unique<Product>(*search->second)};
```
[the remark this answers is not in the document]
oprisor-adrian: The main purpose of the function is to check if there is a product identified by the id in the shopping cart and then return the product so that processing can be done on it. I'll split in two function: _FindProduct()_ that tell me if found a product by _id_ and a _GetProduct()_ method  that return the product.

course_5_homework/main.cc
```diff
@@ -4,38 +4,83 @@
...
+  Store store("MyStore","somewhere",items);
+  // create products and try to insert an empty Product object
+  {
+    std::unique_ptr<Product> prod_01(new Toy("Teddy", "Sc. Toy", 24, 6, 120));
+    std::unique_ptr<Product> prod_02(new Alcoholic("Vodka", 20.99, 21, 750.0, 45));
+    std::unique_ptr<Product> prod_03(new Book("CleanCode", 320, "Rober C. Martin", 340, 5));
+    std::unique_ptr<Product> empty_product(new Product("", 0.0, 0));
+   
+    // initiliaze a store
+    store.AddProduct(prod_01,"cod_prod_1");
```
enachebogdan, line 40: I know this is just an example/homework project, but in general you should add try/catch block for every function that can throw, not just when you know it will fail.

course_5_homework/main.cc
```diff
@@ -4,38 +4,83 @@
...
-  std::unique_ptr<Printer> printer(new ConsolePrinter);
-  printer->Print(store);
-  printer->Print(customer);
+  // create a Customers and make some purchases
+  {
+    std::unique_ptr<Product> prod_01(new Toy("Teddy", "Sc. Toy", 24, 6, 3));
+    std::unique_ptr<Product> prod_02(new Alcoholic("Vodka", 20.99, 21, 750.0, 1));
+    PremiumCustomer premium_customer("PremiumCustomer", "somewhere");
+    
+    premium_customer.AddToCart("cod_prod_1", prod_01);
```
enachebogdan, line 64: AddToCart can throw an exception, you should try/catch

enachebogdan (changes requested):
I've added just a couple more comments, things look good, almost ready to approve.

course_5_homework/src/printer/file_handler.cc
enachebogdan: Actually I can see my comment is not very clear, I meant to ask if there is a need to check if the open mode is read or write, before trying to read or write

course_5_homework/include/customer/customer.h
enachebogdan: Maybe I have missed it, but I haven't seen the returned value used anywhere in your code, so I was thinking maybe it could be a simple void.

oprisor-adrian:
I tried to use forward declaration but encountered errors when I was running. I checked if the class was fully-declared before making the inheritance, but there were still errors. Maybe I missed something . In the future I will try to use forward-declaration.

course_5_homework/include/customer/customer.h
oprisor-adrian: I got an example in _main.cc_. The function _CheckOut()_ was renamed in _CompletePurchase()_.
`try{
      std::cout << "Premium customer has to pay: " << premium_customer.CompletePurchase(store) << "\n";
      premium_customer.CompletePurchase(store); // try to complete purchases process again (shopping cart will be empty)
    }
    catch (std::runtime_error& e){
      Logger::Log(e.what());
    }`

course_5_homework/src/printer/file_handler.cc
oprisor-adrian: I understand now what to do. My last commit fix that.

enachebogdan (approved):
I'm approving your PR, great job with the homework.

PR #44, opened 2023-03-14
tema5 draft

914-Mereu-Bianca:


readWrite/ReadWrite.cpp
```diff
@@ -1,28 +1,29 @@
 #include "ReadWrite.h"
 
-ReadWrite::ReadWrite(Repository &_repository) : repository(_repository)
-{
+ReadWrite::ReadWrite(std::ostream &out_) : out(out_) {
+    if(!out)
+        throw std::runtime_error("Cannot open file");
```
marisvali, line 5: You use ReadWrite to write to console as well, so the error message isn't exactly right.

main.cpp
```diff
@@ -1,19 +1,91 @@
...
+    std::unique_ptr<ReadWrite> file_printer = std::make_unique<ReadWriteStore>(out, store);
+    file_printer->write();
 
-    r.add(Product("salad", 25));
+    std::unique_ptr<ReadWrite> console_p = std::make_unique<ReadWriteStore>(std::cout, store);
+    console_p->write();
 
-    rw.write();
+    std::unique_ptr<ReadWrite> print_c3 = std::make_unique<ReadWriteCustomer>(out, c3);
+    print_c3->write();
```
marisvali, line 90: It would be nice to create a situation where the ReadWrite hierarchy has a point. Why do we want to be able to treat ReadWriteStore and ReadWriteCustomer objects as if they're the same type (ReadWrite)? For products it makes sense, the store wants to handle a collection of products in a uniform way instead of having a separate container for each type of product. Currently the ReadWrite hierarchy doesn't have an obvious point.

store/Store.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    auto it = inventory.find(prod);   
+
+    if (it->second == quantity)
+        inventory.erase(prod);
+    else
+        it->second -= quantity;
+
+    income += money_spent;
+}
```
marisvali, line 80: This method doesn't work unless someone has a pointer to the object originally added in the inventory, or they call Store::getProduct first. Both are not ok, a better design should be possible.

store/Store.cpp
```diff
@@ -0,0 +1,85 @@
...
+        inventory = std::move(s.inventory);
+    }
+    return *this;
+}
+
+Store::~Store() {}
+
+bool Store::existProduct(std::shared_ptr<Product> prod, int quantity)
+{
+    for (auto &[first, second] : this->inventory)
```
marisvali, line 42: This does solve the problem of looking up objects by their address, not their value. But now the map doesn't offer any advantage. It's more efficient to hold everything in a std::vector if you're going to go through it every time you need an object.

store/Store.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        if (*first == *prod)
+            if (second < quantity)
+                throw std::invalid_argument("The store has only " + std::to_string(second) + " pieces of the required product");
+            else
+                return first;
+    }
+    throw std::invalid_argument("The product was not found");
+}
```
marisvali, line 67: This method shouldn't exist. If I want to check that a product is available, I use existProduct. If I want to retrieve a product, I use purchase. If you were to keep the current system with the map, it would be best to just put this logic in Store::purchase.

store/Store.h
```diff
@@ -0,0 +1,37 @@
...
+
+    bool existProduct(std::shared_ptr<Product> prod, int quantity);
+    std::shared_ptr<Product> const getProduct(std::shared_ptr<Product> prod, int quantity);
+    void purchase(std::shared_ptr<Product> prod, int quantity, double money_spent);
+
+    bool operator==(Store const &s) const;
+
+private:
+    double income;
+    std::unordered_map<std::shared_ptr<Product>, int> inventory;
```
marisvali, line 34: This can be made to work well, technically. Cristi even gave an idea about using the KeyEqual in the unordered_map declaration (https://en.cppreference.com/w/cpp/container/unordered_map):
template <
    class Key,
    class T,
    class Hash = [std::hash]
    class KeyEqual = [std::equal_to]
>

However, this design shows a more philosophical flaw. You identify products by the Product class. Each product is identified by its name and price. And if in the future we add producer, weight and size, someone who wants to ask for a "chocolate pudding" must know the price, producer, weight and size. The actual problem is that you're using Product for two purposes: as an identifier for a product type and as a source of information about the product type. These are different things. The store wants to keep information about each product type, that's one thing. The store and the customers want to talk about products using some sort of identifier. That's another thing. The individual products themselves could also be called Product, but that's yet another concept (it could be required if each food item needs an individual expiration date, for example). If you watch the previous presentation there's a more in-depth discussion about exactly this topic.

store/Store.cpp
```diff
@@ -0,0 +1,85 @@
...
+    else
+        it->second -= quantity;
+
+    income += money_spent;
+}
+
+bool Store::operator==(Store const &s) const
+{
+    return income == s.income and inventory == s.inventory;
+}
```
marisvali, line 85: Same as the default implementation, remove it.

store/Store.h
```diff
@@ -0,0 +1,37 @@
...
+class Store
+{
+
+public:
+    Store(int _income, std::unordered_map<std::shared_ptr<Product>, int> _inventory);
+
+    Store(const Store &s);
+    Store(Store &&s) noexcept;
+    Store &operator=(const Store &s);
+    Store &operator=(Store &&s) noexcept;
```
marisvali, line 17: No need for these.

marisvali (changes requested):
Exceptions are used for error handling, which is ok, but this part is missing from the assignment:

"Add at least two cases where you catch an error and try to recover from it."

This means that you should try to catch an exception and recover from it somewhere else than in main. You'll catch and treat everything in main in the end anyway.

store/Store.cpp
marisvali: This function should now take advantage of UnorderedMap, instead of iterating through each item in the map. The find function will find prod correctly now.

unorderedMap/UnorderedMap.h
```diff
@@ -0,0 +1,39 @@
...
+{
+
+public:
+    // UnorderedMap(){};
+    // void add(std::shared_ptr<Product> p, int q) { map.insert(std::make_pair(p, q)); }
+    // bool exist(std::shared_ptr<Product> p) { return map.find(p) != map.end(); }
+
+private:
+    std::unordered_map<std::shared_ptr<Product>, int, ProductHash, ProductEqual> map;
+};
```
marisvali, line 37: Currently none of the products derived from Product have any extra members. But if you ever have a new product that has an extra member, your hash and equal operations will ignore that new member. Consider the following code:
```
class Mouse : public Product 
{
public:
    Mouse(std::string _name, int _price, std::string _color): Product(_name, _price), color(_color) {}
    bool isRestricted() const override { return false; }
    ~Mouse() override {}
protected:
    std::string color;
};

int main()
{
    std::shared_ptr<Product> p1 = std::make_shared<Mouse>("Microsoft Bluetooth® Mobile Mouse 3600", 10, "black");
    std::shared_ptr<Product> p2 = std::make_shared<Mouse>("Microsoft Bluetooth® Mobile Mouse 3600", 10, "white");
    UnorderedMap inventory;
    inventory[p1] = 100;
    Store store = Store(0, inventory);
    
    std::cout << store.existProduct(p1, 1) << "\n";
    std::cout << store.existProduct(p2, 1) << "\n";
    
    std::cout << "\n\n\n\n";
}
```
The store says both white and black variants of the mouse are available. But only the black variant is available. So how do you fix this? You need some kind of virtual comparison function, but it needs to work only between derived objects of the same type. It gets complicated quick.

I admire the commitment to making the design work with map<Product, int>. Currently it still does not address the fact that a customer cannot ask if 'chocolate pudding' is available. He must ask for 'chocolate pudding that costs 10' is available, which is almost never what you want.

Think of what you can do to fix it.

customer/Customer.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    try
+    {
+        s.purchase(p, quantity, p->getPrice() * quantity);
+        shopping_cart[p] += quantity;
+    }
+    catch (std::invalid_argument &e)
+    {
+        std::cout << e.what() << '\n';
+    }
```
marisvali, line 40: It doesn't make sense for the customer to eat the exception here. The buy failed, but the function that calls buyProduct never finds out. And the buyProduct function doesn't do anything to try recover from the error.

customer/Customer.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    try
+    {
+        s.purchase(p, quantity, p->getPrice() * quantity * (1 - discount));
+        shopping_cart[p] += quantity;
+    }
+    catch (std::invalid_argument &e)
+    {
+        std::cout << e.what() << '\n';
+    }
```
marisvali, line 74: Same as with RegularCustomer.

customer/Customer.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    try
+    {
+        s.purchase(p, quantity, p->getPrice() * quantity);
+        shopping_cart[p] += quantity;
+    }
+    catch (std::invalid_argument &e)
+    {
+        std::cout << e.what() << '\n';
+    }
```
marisvali, line 113: Same as with RegularCustomer.

main.cpp
```diff
@@ -1,19 +1,104 @@
...
+        if (type == "premium")
+            customer = std::make_unique<PremiumCustomer>(name, age);
+        if (type == "underage")
+            customer = std::make_unique<UnderageCustomer>(name, age);
+        return customer;
+    }
+    catch (std::invalid_argument &e)
+    {
+        std::cout << e.what() << '\n';
+    }
```
marisvali, line 25: In case of an error, it returns a nullptr and the caller has to check the return result otherwise it generates a segmentation fault (trying to dereference a null pointer). There's no obvious value to catching here.

marisvali (changes requested):
Some previous comments have been addressed, some still need to be addressed. Here are the comments for the new additions.

main.cpp
```diff
@@ -5,56 +5,91 @@
...
+        customer = std::make_shared<UnderageCustomer>(name, age, budget);
+        std::cout << e.what() << '\n';
+    }
+
+    try
+    {
         if (type == "underage")
-            customer = std::make_unique<UnderageCustomer>(name, age);
+            customer = std::make_shared<UnderageCustomer>(name, age, budget);
         return customer;
```
marisvali, line 34: This line isn't needed now.

marisvali (commented):
Great, it looks pretty good so far. Just a few minor comments. And there's a previous comment still unaddressed. Once you address all these, I'll approve. I consider the homework done at this point.

main.cpp
914-Mereu-Bianca: I think I need this part for the case in which I create an Underage customer with createCustomer("c2", 13, 100, "underage")

marisvali: I think I missclicked, I was talking about 'return customer;'

marisvali (approved):
Thanks for the commit. Cristi pointed out that we should be using 'using =' instead of typedef, which is true. I'll approve the pull request for now and you can change it another time.

PR #46, opened 2023-03-21
Upload homework 6

oprisor-adrian:
TO DO:
- [x] code refactoring
- [x] integrating google test
- [x] writing unit tests
- [x] fix bugs


marisvali (changes requested):
Looking through the changes it seems like you refactored the code a lot. Normally this makes sense, but for this homework you shouldn't do that. The purpose for this homework is to keep as much as you can from the original author, make only the changes needed to fix bugs. This is in order to simulate the job experience. When you work with someone else's code at a job, you will often not like their style or solutions, or have good reasons to disagree with their entire approach. However, most of the time you cannot refactor things. Either you are not given enough time or enough permissions. So for this homework we want to simulate this situation where you have to adapt as much as possible to the original code. Unfortunately this means many of the changes you made have to be reverted and only bugfixes kept.

include/shared.h
```diff
@@ -11,12 +11,13 @@ class shared_ptr {
         shared_ptr(T* pointer_object); // constructor
         shared_ptr(const shared_ptr& other); // constructor de copiere
         shared_ptr(shared_ptr&& other); // move construct
-        shared_ptr& operator=(const shared_ptr& other); // copy assignment
+        template<class Y> friend class shared_ptr;
```
enachebogdan, line 14: Interesting idea to access private data of "other" with a different type (Y) than the current class template(T).
Do you think there is another way around this problem, without using friend?

include/shared.h
```diff
@@ -75,32 +76,21 @@ T* shared_ptr<T>::operator->() const {
...
-    (*count)--;
-
-    if (*count == 0) { // daca nu mai am obiecte, eliberez memoria
-        delete data;
-        delete count;
-    }
-    else { // altfel eliberez doar pointer-ul data obiectului curent
-        data = nullptr;
-    }
+    this->reset();
```
enachebogdan, line 79: Did you encounter any error with the old logic? What was the problem/fix

include/shared.h
```diff
@@ -122,23 +112,24 @@ shared_ptr<T>& shared_ptr<T>::operator=(shared_ptr&& other) {
...
-    else { // altfel eliberez pointer-ul data obiectului curent
-        data = nullptr;
+    if ((*count) == 0) {
+      delete data;
+      delete count;
+      data = nullptr;
+      count = nullptr;
+    } else {
+      data = nullptr;
+      count = new int(1);
```
enachebogdan, line 130: Same question here, please describe what was wrong with the old version and how you fixed it.

src/store.cc
```diff
@@ -7,7 +7,7 @@ Inventory::iterator Store::FindInStore(shared_ptr<Product> product)
 {
   for (auto result = inventory_.begin(); result != inventory_.end(); result++)
   {
-    if (result->first == product)
+    if ((*result->first) == *product)
```
enachebogdan, line 10: Same question here, please explain what was the problem with comparing the pointers instead of comparing the objects pointed to.

tests/store_test.cc
```diff
@@ -0,0 +1,7 @@
+#include <gtest/gtest.h>
+#include "../include/store.h"
+
+TEST(GetBalance, TesFuntion){
+  Store store("Store name", 124);
+  EXPECT_EQ(store.GetBalance(), 124);
```
enachebogdan, line 6: Good job integrating the unit test lib. I hope you plan adding tests for all the functions required in the homework.
I see you've added a test file just for store.cpp, which is good. It's encouraged to have unit test for different classes in different files.

enachebogdan (commented):
Code looks good so far, I've added some questions just to check how you fixed some issues. I'll have another look once you add more unit tests.

include/shared.h
oprisor-adrian: First I added an _if_ statement before decrementing the _counter_ to check if a shared_ptr exists and to prevent the decrementation of nullptr. Then I added an initialization to _count_ because after calling reset the current shared_ptr no longer shares ownership of the managed object. So he is the only shared_ptr object managing the null pointer.

include/shared.h
oprisor-adrian: Same explanation as at _reset()_ method.

include/shared.h
oprisor-adrian: Yep. I can use _use_count()_ to access the counter and _*other_ to access the managed object of _shared_ptr<Y>& other_ to create a copy.
I didn't see the _use_count()_ implementation earlier.

include/shared.h
```diff
@@ -75,32 +89,34 @@ T* shared_ptr<T>::operator->() const {
...
 template <class T>
-shared_ptr<T>& shared_ptr<T>::operator=(const shared_ptr& other) {
-    
-    if (this != &other && this->data != nullptr) {
-        this->reset();
+template <class Y>
+shared_ptr<T>& shared_ptr<T>::operator=(const shared_ptr<Y>& other) {
+  if(this->get() != other.get()){
+    reset();
+    data = new T(*other);
```
enachebogdan, line 100: what happens if *other is nullptr?

include/unique.h
```diff
@@ -24,7 +24,7 @@ class unic_ptr {
         T* operator->();
         ~unic_ptr(); // destructor
         /* Observers */
-        const unic_ptr& get() const; // getter
+        T* get() const; // getter
```
enachebogdan, line 27: should you return const pointer?

enachebogdan (approved):
Great job Adrian, I had a look at the tests and they look good overall, hope I didn't miss anything.
Just added 2 small questions, but nothing that important, so I'm approving your PR, homework is considered done.

include/shared.h
oprisor-adrian: I think that would result in an error because _data = new T(*other)_ will attempt to dereference a null pointer.
This can be solved by checking if _other_ is different by null pointer before the copied to be done.

include/unique.h
oprisor-adrian: Const pointer would prevent _unic_ptr_ managed resources from modifies. That is good if I want to provide a way to access the resource in read-only mode. Also if I have a function that modifies the state of the unique pointer, can no longer be called because of const. 
In my project _get()_ method can return a const pointer because I don't modify the state anywhere. 
As a general rule, If I can say it like that and after I google it a bit, the _get_ method should return a non-const pointer to managed object to allow the user to modify the object. On the other side, if the smart pointer is const, _get_ should return a const pointer.

enachebogdan: I agree with you, usually if a class holds a pointer I would not advise exposing it non-const outside of the class (or exposing a non-const reference to an internal member). But since our class is a smart pointer, we don't want it to restrict the access in any way, we just want to take advantage of RAII design.

PR #48, opened 2023-03-22
hw-testing

914-Mereu-Bianca:


product.hpp
```diff
@@ -6,7 +6,6 @@
 class Product {
```
enachebogdan, line 6: You have 2 pair of files, product.hpp/product.cpp, and productS.hpp/productS.cpp. They both define a "Product" class. You should check which one you use and get rid of the other.

store.cpp
```diff
@@ -34,7 +34,7 @@ void StorePrinter::PrintInventory() const
 
     for (const auto& inv : store_.GetProducts())
     {
-        if (inv != nullptr)
```
enachebogdan, line 37: Was there a compile error with the old logic? Or runtime error?

test_file.cpp
```diff
@@ -0,0 +1,12 @@
+#include <gtest/gtest.h>
+#include "unique_ptr.h"
+#include "store.hpp"
+
+
+//unique ptr test
+TEST(HelloTest, BasicAssertions) {
+
+  UniquePointer<int> p = UniquePointer<int>(new int(10));
```
enachebogdan, line 9: It's great that you managed to include the tests in your program, good job. I look forward to checking the rest of the unit tests once you'll add them.
Just a tip, usually it's better to add unit test cpp files for each class, don't add them all in one class. (for your projects, it doesn't really matter if you have 20-30 tests in a single file, but for huge projects it will be hard to manage/search).

unique_ptr.h
```diff
@@ -10,11 +10,16 @@ template<class data_type, class deleter_type = std::default_delete<data_type>>
 class UniquePointer{
   public:
     UniquePointer();
-    explicit UniquePointer(data_type* data);
```
enachebogdan, line 13: What was the problem with explicit here?

customer.hpp
```diff
@@ -12,6 +12,7 @@ class Customer {
         void SpendMoney(float sum);
         void RollbackSpendMoney(float sum);
         virtual float GetPriceAfterDiscount(float sum);
+        virtual ~Customer() {};
```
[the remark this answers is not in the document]
914-Mereu-Bianca: There was no problem but when I tried solving the error related to the unique pointer constructor I thought the problem might be here and I didn't rolled back the changes after I found the problem, the same happend to 'explicit' below

store.cpp
914-Mereu-Bianca: It was a compile error because the operator != was not defined in the UniquePointer class but I solved it by comparing the pointer inside the object 'inv' with nullptr by calling the function inv.Get().

tests/test_unique_ptr.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+public:
+  D(int _nr) : nr(_nr){};
+  int getNr() { return nr; };
+
+private:
+  int nr;
+};
+
+TEST(UniquePtrTest, UniquePtrConstructor)
```
enachebogdan, line 22: can you add some more tests for unique pointer? like move constructor for example? and [] data type constructor

enachebogdan (commented):
New code looks good, tests are well structured and written. I assume you are going to add tests for shared pointer as well.

Part of the homework was to debug potential issues with the code, can you respond here with a quick example of what you found/fixed during your investigation?

tests/test_unique_ptr.cpp
914-Mereu-Bianca: Sure, I will add more tests for unique pointer and also for shared pointer

PR #50, opened 2023-03-26
created threads and sync #7

Lys4nder:


homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <condition_variable>
+#include <queue>
+#include <memory>
+#include "products.hpp"
+
+std::mutex mtx;
+std::condition_variable condVar;
+std::queue<std::shared_ptr<Products>> dishQueue;
+
+std::map<std::string, int> inventory = {
```
enachebogdan, line 11: unordered_map is usually quicker than map, do you really need map here?

homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+    for (auto const& [ingredient, quantity] : ingredients) {
+        inventory[ingredient] -= quantity;
+    }
+}
+
+void CreateProduct()
+{
+    while (true)
+    {
+        bool ingredientsAvailable = true;
```
enachebogdan, line 37: it's usually a good idea to declare local variables right where you use them, so in your case just above the "for" loop.

homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+        std::shared_ptr<Products> dish;
+        std::chrono::high_resolution_clock::time_point t1 = std::chrono::high_resolution_clock::now();
+
+        int rndDish = std::rand() % 3;
+        if (rndDish == 0)
```
enachebogdan, line 42: you could use a switch instead of if/else

homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+            {
+                ingredientsAvailable = false;
+                break;
+            }
+        }
+
+        if (!ingredientsAvailable) {
+            std::cout << "Not enough ingredients to create a dish. Stopping production.\n";
+            break;
+        }
```
enachebogdan, line 68: You could have this check inside a different function. CreateProduct() is too large as a function.
Remember the solid principle - S, create function should only create products, not check if ingredients are available.

homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        dish->Prepare();
+        {
+            std::unique_lock<std::mutex> lock(mtx);
+            dishQueue.push(dish);
+            condVar.notify_all();
+            std::chrono::high_resolution_clock::time_point t2 = std::chrono::high_resolution_clock::now();
+            std::chrono::duration<double, std::milli> time_span = t2 - t1;
+            std::cout << "It took " << time_span.count() / 1000 << "s to prepare ";
+            dish->Serve();
```
enachebogdan, line 78: Even if your Serve() function doesn't do anything in this case, I don't think you should call it in the CreateProduct() function.

homework7/main.cpp
```diff
@@ -0,0 +1,121 @@
...
+        if (!ingredientsAvailable) {
+            std::cout << "Not enough ingredients to create a dish. Stopping production.\n";
+            break;
+        }
+
+        dish->Prepare();
+        {
+            std::unique_lock<std::mutex> lock(mtx);
+            dishQueue.push(dish);
+            condVar.notify_all();
```
enachebogdan, line 74: If you have 5 clients waiting for a product, queue is empty, than you create 1 product. Would you want to notify all 5 or notify just one?

homework7/products.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <map>
+
+class Products {
+    public:
+        virtual void Prepare() = 0;
+        virtual void Serve() = 0;
+        virtual std::map<std::string, int> GetIngredients() = 0;
+};
+
+class Pizza : public Products {
```
enachebogdan, line 17: Even if the Pizza/Burger/Pasta classes are small, it's still not a good idea to have them declared/defined in the same .hpp file, each should have it's own file.

Lys4nder:
sure, will improve the solution. it was rather a draft than something final because i didn't have much time. i had to work on and present 3 projects at uni this week..

homework7/burger.hpp
```diff
@@ -0,0 +1,31 @@
+#ifndef burger
```
enachebogdan, line 1: no big deal, but usually we use caps for ifndef defines.

homework7/menu.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef menu
+#define menu
+#include <vector>
+#include <memory>
+#include "pizza.hpp"
```
enachebogdan, line 5: in your header you should include only the headers that you absolutely need here. For example pizza, pasta, burger are not needed here, should be moved to cpp file.
The logic is this: if somebody includes your menu.hpp in 10 different files, each of them will include pizza/pasta/burger hpp files without knowing/needing them.

homework7/products.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <thread>
+#include <random>
+#include <unordered_map>
+
+class Products {
+    public:
+        virtual void Prepare() = 0;
+        virtual std::string GetName() = 0;
+        virtual std::unordered_map<std::string, int> GetIngredients() = 0;
+};
```
enachebogdan, line 15: does your class need a virtual destructor?
What happens if you do this:
Product *p1 = new Pasta();
delete *p1;

homework7/products.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef products
+#define products
+
+#include <iostream>
+#include <chrono>
```
enachebogdan, line 5: Same thing with the includes. Seems easier/cleaner to include everything in the .hpp file instead of every cpp, but it's actually better to include directly in cpp.

homework7/restaurant.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+void Restaurant::ServeProduct(int id)
+{
+    while (!StopFlag)
+    {
+        auto duration = std::chrono::seconds(5 + std::rand() % 11);
+        std::shared_ptr<Products> dish;
+        {
```
enachebogdan, line 63: I am not sure you need this scope, can you please explain why do you use it?

enachebogdan (commented):
Thinks look good, major improvement since your first commit. I added some minor comments, almost ready to go.

homework7/restaurant.cpp
Lys4nder: its true, i dont need it, i created it when i started the homework to test unique_lock and forgot about it :)

homework7/products.hpp
Lys4nder: this happens
error: type ‘class Products’ argument given to ‘delete’, expected pointer
The error occurs because i am trying to delete the object pointed to by p1 by using the dereference operator, which is not good, so i need the virtual destructor

enachebogdan: ah actually my bad, what I was trying to say is this:
if you call delete p1, compiler checks if destructor is virtual. since it's not, it calls the destructor in the base class - ~Product(). It will not call the destructor in derived class, so the data of the object corresponding to the derived class will not be removed, resulting in  a potential memory leak.

enachebogdan (approved):

PR #52, opened 2023-03-27
hw7 threads

914-Mereu-Bianca:


main.cpp
```diff
@@ -1,137 +1,75 @@
...
-std::shared_ptr<Customer> createCustomer(std::string name, int age, int budget, std::string type)
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+#include <thread>
+#include <atomic>
+#include <mutex>
+#include <condition_variable>
+
+std::mutex mutex_dish, mutex_customer;
```
CristianIonutIonescu, line 17: Why do you have the mutex and conditional variable as global? Unless I missed something you use them in restaurant class only. You should declare them as members of it

main.cpp
```diff
@@ -1,137 +1,75 @@
...
-    }
-    catch (const std::invalid_argument &e)
-    {
-        std::cout << e.what() << std::endl;
-    }
-    std::shared_ptr<Product> p3 = std::make_shared<Soda>("orange juce", 15);
-    std::cout << store.existProduct(p1, 1) << " " << store.existProduct(p3, 1) << "\n";
-    c3->buyProduct(store, p3, 5);
-    c3->buyProduct(store, p4, 5);
+    std::thread customers[20];
```
CristianIonutIonescu, line 63: Don't use a c-style array. Use vector or std:: array. Also you have 20 threads but seem to use only 5

restaurant/Restaurant.cpp
```diff
@@ -0,0 +1,43 @@
...
+    int y = rand() % 1000 + 100;
+    dishes.push(std::move(d));
+    std::cout << "The dish " << dishes.back()->getName() << " is being prepared\n";
+    std::this_thread::sleep_for(std::chrono::milliseconds(y));
+    cond_dish.notify_one();
+}
+
+void Restaurant::deliverDish(std::queue<std::shared_ptr<Dish>> &dishes, const std::shared_ptr<Dish> &d, std::mutex &mutex_dish, std::condition_variable &cond_dish)
+{
+    std::unique_lock<std::mutex> mlock(mutex_dish);
```
CristianIonutIonescu, line 30: What's the difference between unique_lock and lock_guard? In this homework I'm ok with using both, but under normal circumstances I would need a good explanation why you use one and not the other. For the sake of exercise, tell me what's the difference between them and when would you use one and not the other

restaurant/Restaurant.cpp
```diff
@@ -0,0 +1,43 @@
...
+    cond_dish.wait(mlock, [&]()
+                   { return dishes.front().get() != nullptr; });
+    std::cout << "The dish " << dishes.front()->getName() << " is being delivered\n";
+    dishes.pop();
+}
+
+void Restaurant::releaseTable(std::mutex &mutex_customer, std::condition_variable &cond_customer)
+{
+    std::lock_guard<std::mutex> guard(mutex_customer);
```
CristianIonutIonescu, line 39: Why do you lock here?

restaurant/Restaurant.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::lock_guard<std::mutex> lock(mutex_dish);
+
+    int y = rand() % 1000 + 100;
+    dishes.push(std::move(d));
+    std::cout << "The dish " << dishes.back()->getName() << " is being prepared\n";
+    std::this_thread::sleep_for(std::chrono::milliseconds(y));
+    cond_dish.notify_one();
+}
+
+void Restaurant::deliverDish(std::queue<std::shared_ptr<Dish>> &dishes, const std::shared_ptr<Dish> &d, std::mutex &mutex_dish, std::condition_variable &cond_dish)
```
CristianIonutIonescu, line 28: Why are you sending dish here? Not seeing it used

CristianIonutIonescu (changes requested):
Overall ok. i have 2 major issues:
1. Move the mutexes and cond variables in restaurant class
2. You're passing way to many shared_ptrs, this looks sketchy at best, has hidden bugs most likely. I would take some time to determine who owns what and if I can pass raw pointers when needed.
Minor issues:
You create only one dish(pasta), please try to create different types of food
Implement prepare methods for your dishes.

Outside the scope of the project, please try to not write over previous homework, it's hard to read.

PS: there might be some coding style issues but due to git diff being hard to read I couldn't tell for sure, will take another look tomorrow

main.cpp
914-Mereu-Bianca: I wrote them there when I started writing the project and I didn't realize I could move them to the restaurant, but I'll move them now

restaurant/Restaurant.cpp
914-Mereu-Bianca: The difference between them is that lock_guard automatically locks a mutex when it's created, and unlocks it when it's destructed and the unique_lock can be use to lock and unlock manually a mutex, it can also be moved and copied. I wanted to try both in this project just to see how they work, but maybe I should've used only lock_guard because I don't need all the features that unique_lock provides.

restaurant/Restaurant.cpp
914-Mereu-Bianca: I wasn't using atomic variable when I wrote this part, but I see now that I don't need it

restaurant/Restaurant.cpp
914-Mereu-Bianca: When I thought about how I should implement the program I wanted each customer to get its specific order, but later when I implemented the thread that creates dishes I realised that the program doesn't really care about what the customer ordered, so I gave up on that idea. I will remove it from there.

CristianIonutIonescu (approved):

PR #53, opened 2023-03-29
HW7 Restaurant Simulation Multithreading

radavid2017:
At first I complicated the project idea, and after some research on the internet, I redesigned the program simpler, focusing on how multithreading works.


tema7DHoka/restaurant.h
```diff
@@ -0,0 +1,27 @@
...
+    DishQueue dish_queue;
+    CustomerQueue customer_queue;
+    std::atomic<int> num_customers;
+
+    // Thread function for creating dishes 
+    // and adding them to the dish queue
+    void chef();
+
+    // Thread function for customers ordering
+    void customer(int id);
```
CristianIonutIonescu, line 24: order is:
methods
members
also, why is num_customers atomic? not seeing it used as atomic so far. Maybe in the next iteration I'll understand

CristianIonutIonescu (commented):
so far ok, waiting for the rest of the code to be able to review it properly

tema7DHoka/dish.h
```diff
@@ -0,0 +1,24 @@
+#ifndef _DISH_
+#define _DISH
+
+#include <string>
+
+class Dish {
+public:
+    virtual void prepare() = 0;
+    virtual std::string name() = 0;
+};
```
enachebogdan, line 10: does your class need a virtual destructor?
What if you do this:
Dish *d1 = new Pizza();
delete d1;

tema7DHoka/dish.h
```diff
@@ -0,0 +1,24 @@
...
+
+class Pizza : public Dish {
```
enachebogdan, line 12: even for these tiny classes, it would be better to have them each in its file.

tema7DHoka/dish_queue.cc
```diff
@@ -0,0 +1,20 @@
...
+#include "dish_queue.h"
+
+void DishQueue::add(std::shared_ptr<Dish> dish) {
+    std::lock_guard<std::mutex> lock(mutex);
+    dishes.push(dish);
+    std::cout << "Dish added to the queue\n";
+    cv_dishes.notify_one();
+}
+
+std::shared_ptr<Dish> DishQueue::remove() {
```
enachebogdan, line 12: since your data type is basically a custom queue, you should name your function "pop" as well. Remove is quite misleading in this context.

tema7DHoka/restaurant.cc
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    // Wait for all the threads to finish
+    chef_thread.join();
+    for (auto& customer_thread : customer_threads) {
+        customer_thread.join();
+    }
+}
+
+void Restaurant::chef() {
```
enachebogdan, line 27: function naming again, it should be a verb, like "cook" "prepare_dishes", something like that.

enachebogdan (changes requested):
Code looks good overall, I had some minor suggestions.
One thing you might have missed, you were supposed to have the threads for restaurant/customers run in an infinite loop, for you it seems it's just a linear run.

tema7DHoka/dish.h
radavid2017: Now that you are saying this, I tested the cases and I see a virtual destructor is a need here. If there is not a virtual destrcutor, the object is either not destroyed, either the program is calling the wrong destructor for the object (the destructor of the base if it has one).

enachebogdan (approved):

PR #54, opened 2023-03-31
Thread HW

aronkis:


H7/src/restaurant.cc
```diff
@@ -0,0 +1,40 @@
...
+    std::cout << "A table has been occupied. There are " << number_of_tables_ - number_of_clients_ << " tables empty.\n";
+}
+
+void Restaurant::MakeDish(Dishes& dishes, const std::shared_ptr<Dish>& dish, std::mutex& dish_mutex)
+{
+    std::lock_guard<std::mutex> lock(dish_mutex);
+    std::cout << dish->GetName() << " is being prepared.\n";
+    dish->Prepare();
+    std::cout << dish->GetName() << " is ready to serve.\n";
+    dishes.push(std::move(dish));
```
enachebogdan, line 20: The way you should use conditional variables, here is the place where you should notify one of the consumers that you created a product.

enachebogdan (commented):
I like the fact that you simulate basically 2 consumer-producer issues.
You simulate the ideal case here, 5 tables 5 customers, so there will be no waiting around.
You should try to have the produce/consume functions built around on a while(true) instruction, forever producing dishes/serving clients.

enachebogdan (approved):

PR #55, opened 2023-04-03
Upload homework 7

oprisor-adrian:


course_7_homework/include/customers/customer.h
```diff
@@ -0,0 +1,29 @@
...
+    bool IsEmpty() const;
+    const std::string& GetName() const;
+    virtual float& GetBalance();
+    virtual void Print();
+  protected:
+    std::vector<std::unique_ptr<Dish>> order_list_;
+    std::string name_;
+    float balance_;
+};
+inline const std::string& Customer::GetName() const{
```
CristianIonutIonescu, line 26: write the function body directly to header

course_7_homework/include/dishes/appetizer.h
```diff
@@ -0,0 +1,20 @@
...
+
+class Appetizer: public Dish{
+  public:
+    Appetizer(const std::string& name, float price, uint32_t cook_time, bool is_vegetarian);
+    bool IsVegetarian() const;
+    void Print() override;
+  private:
+    bool is_vegetarian_;
+};
+inline bool Appetizer::IsVegetarian() const{
```
CristianIonutIonescu, line 17: same as previous comment

course_7_homework/main.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "dishes/dessert.h"
+#include "dishes/main_course.h"
+
+#include "customers/customer.h"
+#include "customers/underage.h"
+
+void createMenu(std::vector<std::unique_ptr<Dish>>& menu){
+  std::unique_ptr<Appetizer> dish_01(new Appetizer("appetizer", 15.6, 12, false));
+  std::unique_ptr<Dessert> dish_02(new Dessert("dessert", 22, 10, "chocolate"));
+  std::unique_ptr<MainCourse> dish_03(new MainCourse("main_course", 37.99, 30, 400));
```
CristianIonutIonescu, line 18: although is correct, I'd like you to make a habit  of using `std::make_unique`.
LE: why not creating them directly in push_back?

course_7_homework/main.cc
```diff
@@ -0,0 +1,76 @@
...
+
+  // add dishes to order list
+  std::vector<std::string> example_order_lists_ = {"appetizer", "dessert", "main_course"};
+
+  // create customers
+  std::vector<std::shared_ptr<Customer>> customers;
+  std::shared_ptr<Customer> customer_01(new Customer("Marius", 250));
+  std::shared_ptr<Customer> customer_02(new Underage("Andrei", 0.1, customer_01));
+  customers.push_back(customer_01);
+  customers.push_back(customer_02);
```
CristianIonutIonescu, line 50: same as with dishes but with `make_shared`

course_7_homework/src/customers/customer.cc
```diff
@@ -0,0 +1,63 @@
...
+void Customer::Order(const std::string& dish_name, Restaurant& restaurant){
+  try{
+    restaurant.AddOrder(dish_name);
+    std::unique_ptr<Dish> dish = restaurant.GetDish(dish_name);
+    if(dish){
+      std::cout << name_ << " ordered " << dish->GetName() << std::endl;
+      order_list_.push_back(std::move(dish));
+    }
+  }
+  catch(std::invalid_argument& e){
```
CristianIonutIonescu, line 19: catch exceptions as const reference

course_7_homework/src/restaurant.cc
```diff
@@ -0,0 +1,103 @@
...
+      if (!is_open_){
+        return;
+      }
+      if(!order_list_.empty()){
+        dish = GetFromMenu(order_list_.front());
+        order_list_.pop();
+      }
+    }
+    if(dish){
+      auto cook_time = std::chrono::milliseconds(dish->GetCookTime());
```
CristianIonutIonescu, line 47: is this used?

course_7_homework/src/restaurant.cc
```diff
@@ -0,0 +1,103 @@
...
+void Restaurant::AddOrder(const std::string& dish_name){
+  for(const auto& dish : menu_){
+    if(dish->GetName() == dish_name){
+      std::unique_lock<std::mutex> lock(dishes_mutex_);
+      order_list_.push(dish_name);
+      lock.unlock();
+      dishes_cv_.notify_one();
+      return;
+    }
+  }
```
CristianIonutIonescu, line 64: search for a STL algorithm that would solve your search without the need to iterate yourself through the whole vector

course_7_homework/include/customers/underage.h
```diff
@@ -0,0 +1,21 @@
...
+
+class Underage: public Customer{
+  public:
+    Underage(const std::string& name, float discount,
+             const std::shared_ptr<Customer>& guardian);
+    float& GetBalance() override;
+    void Pay() override;
+    void Print() override;
+  private:
+    std::shared_ptr<Customer> guardian_;
```
CristianIonutIonescu, line 17: is the underage co-owner of the guardian? If not you can have the guardian as a raw pointer

course_7_homework/include/customers/underage.h
```diff
@@ -0,0 +1,21 @@
...
+    float discount_;
+    float& guardian_balance_;
```
CristianIonutIonescu, line 19: why do you need this when you have the guardian pointer already here?

course_7_homework/src/customers/customer.cc
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void Customer::Pay(){
+  float payment = CalculatePayment();
+  if(balance_ < payment){
+    std::cerr << "error: insuficient balance to pay the order!" << std::endl;
+    return;
+  }
+  balance_ -= payment;
+  order_list_.clear();
```
CristianIonutIonescu, line 39: what happens when you have 2 threads, one for this customer and one for an underage customer that has the guardian the afore mentioned guardian and both need to pay at the same time?
Think of how you can make sure the payment is correct in this case.

course_7_homework/src/restaurant.cc
```diff
@@ -0,0 +1,103 @@
...
+      dishes_cv_.notify_all();
+      std::this_thread::sleep_for(1s);
+    }
+  }
+}
+
+void Restaurant::AddOrder(const std::string& dish_name){
+  for(const auto& dish : menu_){
+    if(dish->GetName() == dish_name){
+      std::unique_lock<std::mutex> lock(dishes_mutex_);
```
CristianIonutIonescu, line 58: why are you using unique_lock here? Please read about `lock_guard` as well. Compare them both and respond back with the differences between them and when do you think you should use one and when the other.

CristianIonutIonescu (commented):
Left a couple of comments. I still have to take a look over the restaurant class and synchronisation mechanism there to make sure everything is alright.

Kudos for adding unit tests to the homework! 👍

CristianIonutIonescu (commented):
Synchronization looks ok. Please fix the other comments

course_7_homework/src/restaurant.cc
oprisor-adrian: I was thinking that I need to explicitly unlock the mutex before notifying the dishes thread. I read about and now I know that the lock_guard will unlock the mutex when he goes out of scope. lock_guard is faster than unique_lock, but unique_lock is more flexible allowing the lock to be released explicitly using the unlock() method. I will change that in the next commit.

course_7_homework/src/restaurant.cc
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+std::unique_ptr<Dish> Restaurant::GetFromMenu(const std::string& dish_name){
+  for(const auto& dish : menu_){
+    if(dish->GetName() == dish_name){
+      return std::make_unique<Dish>(*dish);
+    }
+  }
+  return nullptr;
```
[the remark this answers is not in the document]
oprisor-adrian: I will fix that by creating a new unique_ptr.

course_7_homework/src/customers/customer.cc
oprisor-adrian: Can appear a race tracing. I can fix that by creating a mutex in the Customer class that locks the balance when the guard tries to pay and another mutex in the Underage class that locks the balance when the underage client tries to pay.

CristianIonutIonescu: 2 different mutexes won't solve the issue. You need only one for both cases

course_7_homework/src/customers/underage.cc
```diff
@@ -0,0 +1,35 @@
...
+
+#include <iostream>
+
+Underage::Underage(const std::string& name, float discount, 
+                   const std::shared_ptr<Customer>& guardian)
+    : Customer(name, 0.0), discount_(discount), guardian_(guardian.get()){}
+
+void Underage::Pay(){
+  float payment = CalculatePayment();
+  std::lock_guard<std::mutex> lock(balance_mutex_);
```
CristianIonutIonescu, line 11: you're not locking the correct mutex here. You should lock guardian's mutex not underage mutex

CristianIonutIonescu (commented):
almost done. I'm most interested in fixing the mutex issue you have in underage client.

course_7_homework/src/customers/underage.cc
oprisor-adrian: Right! I'll fix that.

CristianIonutIonescu (approved):

PR #56, opened 2023-04-15
hw8

Lys4nder:


homework8/client.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef  CLIENT
+#define CLIENT
+#include <netinet/in.h>
+
+
+class Client {
+    public:
+        Client(int port, char*ip);
```
CristianIonutIonescu, line 8: don't use char *, use std::string.
Also a good practice is to send the IP first and afterwards the port because under the hood, the address looks like: 127.0.0.1:1502

homework8/server.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef CLIENT
+#define CLIENT
```
CristianIonutIonescu, line 2: ifndef SERVER_HPP?

homework8/client.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef  CLIENT
+#define CLIENT
```
CristianIonutIonescu, line 2: please correct your defines. Extra space in ifndef and missing the _H.
https://google.github.io/styleguide/cppguide.html#The__define_Guard

CristianIonutIonescu (commented):
The review I gave is just a high level overview of things that are not ok.
Aside for the comments I left, as a good practice, I suggest to look at how you could make both client and server to follow RAII.

homework8/client.hpp
Lys4nder: makes sense, i saw that when i tried to curl

CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.22)
-project(tema7)
+project(project)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED true)
-add_executable(tema7 homework7/main.cpp homework7/restaurant.cpp homework7/menu.cpp)
\ No newline at end of file
+add_executable(tema7 homework7/main.cpp homework7/restaurant.cpp homework7/menu.cpp)
+add_executable(server homework8/ServerMain.cpp homework8/server.cpp)
```
enachebogdan, line 6: no big deal, but you could have Client/Server files in separate folder. Imagine they were bigger programs, you'd have to figure out which file is for which.

homework8/client.hpp
```diff
@@ -0,0 +1,26 @@
...
+    public:
+        Client(std::string ip, int port);
+        void CreateSocket();
+        void ConnectToServer();
+        void Close();
+        void Run();
+        void Initialise();
+        ~Client();
+    private:
+        std::unique_ptr<int> client_;
```
enachebogdan, line 17: I can't figure out why you need a pointer for client_, couldn't you use a simple int instead?
Also, client_/server_ are not very suggestive names for sockets, add some "socket" in the name to make it clear what they are.

homework8/server.cpp
```diff
@@ -0,0 +1,110 @@
...
+    while (server_ > 0)
+    {
+        strcpy(buffer_, "Server connection successfull\n");
+        send(*server_, buffer_, 1024, 0);
+        std::cout << "Connected with client.\n";
+        std::cout << "Enter # to end the connection\n";
+        std::cout << "Enter * after every message to send it\n";
+        std::cout << "Client: ";
+        //add switch case to messages and file transfer
+        do {
```
enachebogdan, line 75: Same suggestion here, separate functions for send/receive data

enachebogdan (commented):
Great work Lysander, code looks good overall and gets the job done. I've added some minor comments on top of what Cristi suggested already.

enachebogdan (approved):

PR #57, opened 2023-04-16
Hw8: server-client through socket

radavid2017:


tema8DHoka/client/client.cc
```diff
@@ -0,0 +1,47 @@
...
+    int n = read(socket_file_description, buffer, 255);
+    if (n < 0) {
+        error("Error reading from socket");
+    }
+
+    std::cout << "Received data from server: " << buffer << std::endl;
+}
+
+void Client::close() {
+    ::close(socket_file_description);
```
enachebogdan, line 40: What happens if you create a Client instance, you don't create the socket, then you destroy the instance. You'll call close(...) on an uninitialized socket descriptor. 
A solution that I've seen and suggested elsewhere, you could default-initialize the socket descriptor to a known value and check against that value when you use it.

tema8DHoka/server/server.cc
```diff
@@ -0,0 +1,62 @@
...
+#include <cstdlib>
+#include <cstring>
+#include <unistd.h>
+#include <arpa/inet.h>
+
+Server::Server(int port) {
+    this->port = port;
+}
+
+void Server::start() {
```
enachebogdan, line 12: You should have different functions for creating the socket, bind, listen, etc. Even if you call all of them from a start() function.

tema8DHoka/server/server.cc
```diff
@@ -0,0 +1,62 @@
+#include "server.h"
+#include <iostream>
+#include <cstdlib>
+#include <cstring>
+#include <unistd.h>
+#include <arpa/inet.h>
+
+Server::Server(int port) {
```
enachebogdan, line 8: try using member initializer list by default.

tema8DHoka/client/client.cc
```diff
@@ -0,0 +1,47 @@
...
+}
+
+void Client::error(std::string const& msg) {
+    std::cerr << msg << std::endl;
+    close();
+    exit(1);
```
enachebogdan, line 46: It's a bit extreme that for every error you'd exit the program entirely. For your homework project it might work, but in a more complex context you might have to do more error handling.(nothing to change here, just to keep in mind).

enachebogdan (commented):
Great stuff David, code looks good overall and gets the job done. I added just some minor comments.

tema8DHoka/client/client.cc
radavid2017: Thank you for this warning, I think it is a good idea if I default initialize the descriptor with "NULL" value, and depending on this value I will close if needed.

tema8DHoka/client/client.cc
radavid2017: Thank you for giving me this wider and wiser perspective of view! I hope to remember this advice for the future if I will need in any case.

enachebogdan (approved):

PR #58, opened 2023-04-19
sockets HW

914-Mereu-Bianca:


client/Client.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (connect(socket_descriptor, (struct sockaddr *)&server_addr, sizeof(server_addr)) < 0)
+    {
+        perror("Error connecting to socket");
+        exit(EXIT_FAILURE);
+    }
+}
+
+void Client::sendData(const char *data, int length)
+{
+    send(socket_descriptor, data, length, 0);
```
enachebogdan, line 45: Maybe you could error-check to see if the send was successful.

client/Client.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+void Client::receiveData(char *buffer, int buffer_size)
+{
+    int val = read(socket_descriptor, buffer, buffer_size);
+    printf("%s\n", buffer);
+}
+
+Client::~Client()
+{
+    close(socket_descriptor);
```
enachebogdan, line 56: What happens if you create a Client instance, you don't call CreateSocket(), you just destroy the instance. You'd call close(...) on an uninitialized socket_decriptor.
An elegant solution would be to default-initialize socket_descriptor to a known value, and compare against it when you want to do operations on the socket.

client/Client.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <string.h>
+
+Client::Client()
+{
+}
+
+void Client::createSocket()
+{
+    char buffer[1024] = {0};
+    const char *data_to_send = "Hello from client!";
```
enachebogdan, line 15: you don't need buffer/data here, probably forgot them from some tests or something.

client/Client.cpp
```diff
@@ -0,0 +1,57 @@
...
+    server_addr.sin_port = htons(8080);
+
+    if (inet_pton(AF_INET, "127.0.0.1", &server_addr.sin_addr) <= 0)
+    {
+        perror("inet_pton failed");
+        exit(EXIT_FAILURE);
+    }
+
+    // Connect to the server
+    if (connect(socket_descriptor, (struct sockaddr *)&server_addr, sizeof(server_addr)) < 0)
```
enachebogdan, line 36: I'd separate the logic for connecting to server into a different function. CreateSocket should only do one thing.

server/Server.cpp
```diff
@@ -0,0 +1,87 @@
+#include "Server.h"
+#include <unistd.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <string>
+#include <string.h>
+
+Server::Server(int port_)
```
enachebogdan, line 9: use member initializing list anytime it's possible.

enachebogdan (commented):
Great job Bianca, code looks good, I only had minor comments.
I find it a bit aggressive to use exit() at almost any failure on client/server, it makes sense maybe for your app but in other context you might have to do some more error handling. No need to change this for the homework, just to keep it mind.

enachebogdan (approved):

PR #59, opened 2023-04-19
Upload homework 8

oprisor-adrian:


course_8_homework/include/app.h
```diff
@@ -0,0 +1,16 @@
+#ifndef SOCKET_APPLICATION_INCLUDE_APP_H_
+#define SOCKET_APPLICATION_INCLUDE_APP_H_
+
+#include <string>
+
+#include "connection.h"
```
enachebogdan, line 6: can you forward declare instead?

course_8_homework/src/app/app.cc
```diff
@@ -0,0 +1,31 @@
...
+    std::cerr << e.what() << std::endl;
+  }
+}
+
+void App::SendMessage(const std::string& message){
+  try{
+    connection_.Send(message);
+    std::cout << "[ app info ]: Sent data to server: " << message << std::endl;
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
+    std::string message = connection_.Receive();
```
enachebogdan, line 24: Your SendMessage fuction does 2 things: sends and receives, which is no ideal. Can you split the receive logic into another function?

course_8_homework/include/connection.h
```diff
@@ -0,0 +1,27 @@
+#ifndef SOCKET_APPLICATION_INCLUDE_CONNECTION_H_
+#define SOCKET_APPLICATION_INCLUDE_CONNECTION_H_
+
+#include <string>
+
+#include <netinet/in.h>
+
+class Connection{
```
enachebogdan, line 8: You have both server/client logic in this class, it might be confusing for somebody who's not familiar with your implementation, when trying to decide what functions to use. And it's not very compliant with "S" SOLID principle.
You could have a base class with functionality common to both, and 2 derived classes for Client/Server specific logic.

enachebogdan (commented):
Good work Adrian, code is pretty clean and gets the job done. I added few small comments, nothing important anyway.

course_8_homework/include/connection.h
oprisor-adrian: Indeed, the Connection class doesn't exactly comply with the "S" of the SOLID principle. I'll refactor the code and find a better solution.

course_8_homework/include/connection.h
```diff
@@ -0,0 +1,22 @@
...
+  public:
+    Connection(const std::string& hostname, const int port);
+    ~Connection();
```
enachebogdan, line 11: you will not have any issues because you don't actually use polymorphism in your homework, but keep in mind to declare your destructors virtual when you design base classes.

enachebogdan (approved):

enachebogdan (approved):

PR #60, opened 2023-04-20
ClientServer first commit.

aronkis:


enachebogdan (commented):
Great stuff Aron, code is well written, clean and gets the job done.
One thing that I have to complain about, you don't seem to ever close your sockets, that will lead to a leak. Maybe you can make those destructors useful.

enachebogdan (approved):

PR #61, opened 2023-04-24
Upload homework 9

oprisor-adrian:
To do:
- [x] Multithread and thread pool
- [x] Object serialization
- [x] Server logs
- [x] Bash script


enachebogdan (commented):
Great Job Adrian, your code looks really good. I'm happy that you integrated thread pool in the homework. Thread sync, sockets logic, serialization, script, all look good.

No need to change this, but I feel my suggestion from last homework, to have a base class "connection", might have pushed you to a design that's a bit ambiguous for somebody not familiar with your code; you should have a client and a server class performing all the operation(instead of client_connection/server_connection). They should have a member connection, instead of inheriting from it. I'd rather have some duplicate code than not being able to understand what's the exact purpose of each class.

Another thing that you might consider, server/client are 2 separate apps, they shouldn't really have common include/src folders. (again, no need to change anything for now).

enachebogdan (approved):

PR #62, opened 2023-04-25
hw9

Lys4nder:


homework9/Message.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef MESSAGE_HPP
+#define MESSAGE_HPP
+
+#include <boost/archive/text_oarchive.hpp>
+#include <boost/archive/text_iarchive.hpp>
+#include <boost/serialization/string.hpp>
+#include <boost/serialization/vector.hpp>
```
enachebogdan, line 7: are you sure you need all of these includes from boost in your header file? maybe you can move them in .cpp instead.
Also, I don't think you need vector and iostream here as well.

homework9/Message.hpp
```diff
@@ -0,0 +1,36 @@
...
+private:
+    friend class boost::serialization::access;
+    template<class Archive>
+    void serialize(Archive& ar, const unsigned int version)
+    {
+        ar & message_;
+        ar & sender_;
+        ar & timeStamp_;
+    }
+    std::string message_;
```
enachebogdan, line 30: It's usually a bad idea to name a member the same as the class, it's confusing for somebody reading into your code. Could rename it to "text" or something like that.

homework9/Message.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <vector>
+
+class Message {
+public:
+    Message() = default;
+    Message(std::string message, std::string sender, std::string timeStamp);
+    inline void SetMessage(std::string message) {message_ = message;}
+    inline void SetTimeStamp(std::string timeStamp) {timeStamp_ = timeStamp;}
+    inline void SetSender(std::string sender) {sender_ = sender;};
+    void LogMessage();
```
enachebogdan, line 19: No need to change anything for this homework, but I'm not a huge fan of Message class logging itself.
The way I see it, it is the Server who needs to log data, so the Message could just have a "GetData" or something like that to help the log process.

homework9/Message.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+void Message::LogMessage()
+{
+    time_t now = std::time(nullptr);
+    char timestamp[20];
+    std::strftime(timestamp, sizeof(timestamp), "%Y-%m-%d %H:%M:%S", std::localtime(&now));
+    
+    Message logMessage(message_, sender_, timestamp);
+
+    std::stringstream ss;
```
enachebogdan, line 16: It's really not important to serialize a message in a log file, there's no benefit to do it. You only need to serialize when you want to load objects back in your program when you must deserialize them.

homework9/Message.cpp
```diff
@@ -0,0 +1,26 @@
...
+Message::Message(std::string message, std::string sender, std::string timeStamp) 
+    : message_(message), sender_(sender), timeStamp_(timeStamp) {}
+
+void Message::LogMessage()
+{
+    time_t now = std::time(nullptr);
+    char timestamp[20];
+    std::strftime(timestamp, sizeof(timestamp), "%Y-%m-%d %H:%M:%S", std::localtime(&now));
+    
+    Message logMessage(message_, sender_, timestamp);
```
enachebogdan, line 14: creating an instance of Message when you are already in an instance of the class doesn't make much sense for me. Imagine your server processes(and logs) 1000 messages per second. It would be really inefficient creating these extra instances.
You could try setting the timestamp into the current message, then use *this for serialization.

homework9/server/server.cpp
```diff
@@ -0,0 +1,122 @@
...
+            msg.LogMessage();
+            send(serverSocket_, buffer_, 1024, 0);
+            if (*buffer_ == '#')
+            {
+                send(serverSocket_, buffer_, 1024, 0);
+                *buffer_ = '*';
+                end_ = true;
+            }
+        } while (*buffer_ != '*');
+        ReceiveMessage();
```
enachebogdan, line 102: You shouldn't receive message in a send message function. SendMessage() should do only one thing, just send the message to client. It shouldn't even log to file.

homework9/client/client.cpp
```diff
@@ -0,0 +1,95 @@
...
+    CreateSocket();
+    ConnectToServer();
+}
+
+void Client::SendMessage()
+{
+    std::cout << "\nClient: ";
+    do {
+        std::cin >> buffer_;
+        send(clientSocket_, buffer_, 1024, 0);
```
enachebogdan, line 64: I think you've missed a bit the homework assignment of serialization. The entire purpose was to serialize a Message object in the client class, send it to Server, than deserialize it in the Server class(optional, do it both ways). So maybe instead of serializing for writing in the file, you could try to do this.

homework9/server/server.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+    size = sizeof(serverAddress_);
+    std::cout << "Looking for client...\n";
+    listen(clientSocket_, 1);
+    serverSocket_ = accept(clientSocket_, (struct sockaddr*)&serverAddress_, &size);
+
+    if (serverSocket_ < 0)
+    {
+        perror("Accepting client failed");
+        exit(EXIT_FAILURE);
```
enachebogdan, line 54: You shouldn't force close the entire Server app if you fail to communicate with one client. Imagine you have 50 clients connected simultaneously: one of them fails, you shut down the Server.

homework9/server/server.cpp
```diff
@@ -0,0 +1,122 @@
...
+{}
+
+Server::~Server()
+{
+    CloseServer();
+}
+
+void Server::CreateSocket()
+{
+    clientSocket_ = socket(AF_INET, SOCK_STREAM, 0); //TCP
```
enachebogdan, line 23: I think you are using clientSocket_ and serverSocket_ naming backwards. This is the server socket that you are creating/binding in the server app, and the client sockets are the ones connecting to it, not the other way around.

homework9/client/client.hpp
```diff
@@ -0,0 +1,27 @@
...
+        void ConnectToServer();
+        void Close();
+        void Run();
+        void Initialise();
+        void SendMessage();
+        void ReceiveMessage();
+        ~Client();
+    private:
+        int clientSocket_ = 8080;
+        int serverSocket_;
```
enachebogdan, line 19: do you ever use this in the Client class? maybe I have missed it. If not, get rid of it please

enachebogdan (commented):
Good job Lysander, things look good so far. I'll have another look after you add threads and serialize the messages you send between server/client.

homework9/Message.cpp
Lys4nder: so should i remove this?

homework9/server/server.cpp
Lys4nder: if i remove that Receive, it wont work properly
image

PR #64, opened 2023-04-26
hw9 

914-Mereu-Bianca:


client/Client.h
```diff
@@ -10,8 +11,7 @@ class Client
     Client();
     void createSocket();
     void connectToServer();
-    void sendData(const char *data, int length);
-    void receiveData(char *buffer, int buffer_size);
+    void sendMessage(Message msg);
```
enachebogdan, line 14: can't you use const ref Message instead? And forward declaration should work fine after you do that.

server/Server.cpp
```diff
@@ -40,44 +49,47 @@ void Server::bindServer()
...
+    while (true)
     {
-        perror("accept failed");
-        exit(EXIT_FAILURE);
+        int server_len = sizeof(server_addr);
+        new_socket = accept(server_descriptor, (struct sockaddr *)&server_addr, (socklen_t *)&server_len);
+        if (new_socket <= 0)
+        {
+            perror("accept failed");
+            exit(EXIT_FAILURE);
```
enachebogdan, line 68: I would advise you to handle the error in a safer manner. If one client can't connect, you shouldn't close a server app for all other clients.

server/Server.cpp
```diff
@@ -40,44 +49,47 @@ void Server::bindServer()
...
-    int val = read(new_socket, buffer, 1024);
-    if (val < 0)
+    char buffer[1024];
+
+    int bytes_received = read(client_socket, buffer, sizeof(buffer));
+    if (bytes_received < 0)
     {
-        perror("read failed");
+        perror("Error receiving data from client.");
         exit(EXIT_FAILURE);
```
enachebogdan, line 82: same advise, get rid of the exit

server/Server.cpp
```diff
@@ -5,9 +5,18 @@
...
+#include <sstream>
+#include <../../boost/serialization/string.hpp>
+#include <../../boost/archive/text_iarchive.hpp>
+#include <../../boost/serialization/export.hpp>
+#include <../../boost/serialization/access.hpp>
+#include "../message/Message.h"
 
 Server::Server(int port_) : port(port_) {}
 
+std::ofstream Server::logfile("../message/Message.txt", std::ios_base::out | std::ios_base::trunc);
```
enachebogdan, line 18: you should move the definition of static members before all the function definitions(so above the constructor in your case).
Is there a particular reason why you decided to use static? Nothing wrong with it, anyway, but I think a regular class member would have worked as well.

server/Server.cpp
```diff
@@ -86,4 +98,11 @@ Server::~Server()
         close(new_socket);
     if (server_descriptor != -1)
         close(server_descriptor);
+
+    logfile.close();
+
+    for (int i = 0; i < clients.size(); i++)
```
enachebogdan, line 104: let's use a modern way to iterate instead of i++

enachebogdan (commented):
Nice work Bianca, code looks good, I had just some minor comments.

server/Server.cpp
914-Mereu-Bianca: I considered that I would use the same txt file for now for any server so I made the logfile static

enachebogdan (approved):

PR #65, opened 2023-04-28
Last one.

aronkis:


H9/src/server.cc
```diff
@@ -0,0 +1,96 @@
...
+
+void Server::HandleClient(int socket)
+{
+  char buffer[1024];
+
+  int bytes_received = read(socket, buffer, sizeof(buffer));
+  if (bytes_received < 0)
+  {
+    perror("ERROR WHILE HANDLING A CLIENT!\n");
+    exit(EXIT_FAILURE);
```
enachebogdan, line 68: Usually it's not a good idea to force close the server if you can't communicate to one client. Imagine you have 50 clients connected, if there's an error with one of them you don't want to close the server for everyone else.
I've seen this approach with "exit" in almost all the homeworks, and works fine for small projects, but in other context we would need proper error/exception handling.(no need to change for this homework).

H9/src/client.cc
```diff
@@ -0,0 +1,45 @@
...
+void Client::ConnectToServer()
+{
+  if (connect(socket_info_, (struct sockaddr*)& server_addr_, sizeof(server_addr_)) < 0)
+  {
+    perror("ERROR WHILE CONNECTING!\n");
+    exit(EXIT_FAILURE);
+  }
+}
+
+void Client::SendData(const Message message)
```
enachebogdan, line 34: use const Message& instead.

H9/src/server.cc
```diff
@@ -0,0 +1,96 @@
...
+  {
+    int server_len = sizeof(server_addr_);
+    new_socket_ = accept(server_info_, (struct sockaddr *)&server_addr_, (socklen_t *)&server_len);
+
+    if (new_socket_ < 0)
+    {
+      perror("ERROR WHILE ACCEPTING A CLIENT!\n");
+      exit(EXIT_FAILURE);
+    }
+    clients_.emplace_back(std::thread(Server::HandleClient, new_socket_));
```
enachebogdan, line 94: You are not joining the threads anywhere.

enachebogdan (commented):
Good job Aron, code looks clean and simple, gets the job done. I only added some minor comments. 
And I'm not sure about being the "Last one" though 😃

aronkis:
> Good job Aron, code looks clean and simple, gets the job done. I only added some minor comments. And I'm not sure about being the "Last one" though 😃

Last one meant as last homework. :D

